PR #5, opened 2015-07-22
Fix the demo LRS endpoint URL display

mudrd8mz:
The trailing quote was considered as a part of the URL when viewed via the Github interface.


jlowe64:
Thank you! I have merged this back into the repo.

PR #21, opened 2015-09-11
Fix assignment submission not emitting statement

micaherne:
Hi, I'm finding that when I submit an assignment in Moodle, a statement isn't being sent to the LRS. The attached patch should fix this.

I think what is happening is that the \mod_assign\event\assessable_submitted event is being passed to insert_event_entries() in a list containing other events that aren't currently handled (specifically \assignsubmission_file\event\assessable_uploaded). This is causing a null parameter to be passed to $translatorcontroller->createEvent() which isn't allowed by the function signature, so it's throwing an exception which bails out of insert_event_entries() before the assessable_submitted event can be handled.

### DIFF
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -63,7 +63,7 @@ class store extends php_obj implements log_writer {
      * Should the event be ignored (not logged)? Overrides helper_writer.
      * @param event_base $event
      * @return bool
-     * 
+     *
      */
     protected function is_event_ignored(event_base $event) {
         return false;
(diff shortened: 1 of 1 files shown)


jlowe64:
Thanks for bringing this to my attention! I'll look into it next week.

jlowe64:
Sorry, this didn't happen last week. My organisation was moving infrastructures and it took my whole week! I'll be trying to get this, or something similar, into a release by end of week.

micaherne:
No problem, thanks for looking at it!

jlowe64:
@micaherne would it not be better to perhaps do it this way:

```
if (!is_null($moodleevent)) {
    $translatorevent = $translatorcontroller->createEvent($moodleevent);
}
```

etc... and the same for the other? So instead of catching the nulls and routing them to a dummy function, you just catch the !is_nulls and process those?

micaherne:
Sorry, I realised my patch had the continue statements commented out by mistake (which I guess might have made it look like some kind of anonymous function, but it isn't). I've fixed it now.

I suppose you could do it the way you're suggesting although it would be more lines of code - as you'd need to initialise the variables to null to avoid "undefined variable" messages -  and less efficient, since you'd still be checking the second one even if you already know it will be null. That said, it doesn't really matter how it's implemented, as long as the exceptions stop being thrown!

jlowe64:
I see those continues in the other PR. Perhaps we can just focus on that one, get it working, and go ahead with that. I can see how with larger organisations they may need to schedule the cron out a bit more.

jlowe64:
Actually, I'm going to put this in 0.4.0, and delay the other PR for 0.5.0 until we can make sure it works.

micaherne:
Thanks for pulling this in. Seems like a good idea to do bugfixes and new features separately.

I don't know how this patch ended up in my other pull request too - I'm usually quite careful about these things! When I get a chance I'll rebase the other one to avoid any problems.

PR #26, opened 2015-09-18
Background processing mode

micaherne:
Hi, I wonder if you'd be interested in this patch which enables statements to be sent to the LRS as a scheduled task, rather than as a blocking HTTP request during user time?

We created it because will be using the plugin for a learning analytics project and had some concerns that the performance of our Moodle service would be tied to the responsiveness of the LRS. In our case, the LRS is run by an external body and, while we're confident they will provide a robust service, it would not be sensible from a service provision position to rely on this.

We have had very bad experiences with this kind of thing before with a plagiarism plugin that caused a certain important page on our Moodle server to grind to a halt if the service was down or in maintenance, and the effects from the logstore_xapi plugin would be much, much worse given that it is firing on every login, course view and most module views.

### DIFF
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -63,7 +63,7 @@ class store extends php_obj implements log_writer {
      * Should the event be ignored (not logged)? Overrides helper_writer.
      * @param event_base $event
      * @return bool
-     * 
+     *
      */
     protected function is_event_ignored(event_base $event) {
         return false;
@@ -75,6 +75,19 @@ class store extends php_obj implements log_writer {
      *
      */
     protected function insert_event_entries(array $events) {
+        global $DB;
+
+        // If in background mode, just save them in the database
+        if (get_config('logstore_xapi', 'backgroundmode')) {
+            $DB->insert_records('logstore_xapi_log', $events);
+        } else {
+            $this->process_events($events);
+        }
+
+    }
+
+    public function process_events(array $events) {
+
         // Initializes required services.
         $xapicontroller = new xapi_controller($this->connect_xapi_repository());
         $moodlecontroller = new moodle_controller($this->connect_moodle_repository());
@@ -82,11 +95,20 @@ class store extends php_obj implements log_writer {
 
         // Emits events to other APIs.
         foreach ($events as $event) {
+            $event = (array) $event;
             // $this->error_log('');
             // $this->error_log_value('event', $event);
             $moodleevent = $moodlecontroller->createEvent($event);
+            if (is_null($moodleevent)) {
+                continue;
+            }
+
             // $this->error_log_value('moodleevent', $moodleevent);
             $translatorevent = $translatorcontroller->createEvent($moodleevent);
+            if (is_null($translatorevent)) {
+                continue;
+            }
+
             // $this->error_log_value('translatorevent', $translatorevent);
             $xapievent = $xapicontroller->createEvent($translatorevent);
             // $this->error_log_value('xapievent', $xapievent);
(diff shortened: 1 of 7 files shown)


garemoko:
I haven't looked at the code or tested this, but it sounds like a positive change. Is there any reason this should be a setting rather than just how the plugin behaves?

micaherne:
I guess the only downside is that you don't get "real time" data in the LRS, but I can't really imagine a scenario where that would be so crucial that someone would be prepared to take the performance hit on their server for it. By doing it this way, you're also accepting another database write (for the same event data that's also going into the standard log) but I'd take that over an HTTP request over the web any day!

From my point of view, I think the plugin would be much better if it worked in the background by default, but I did the patch this way so it was backwards compatible. I'd be happy to rework it to work this way by default.

The other useful aspect of this that I didn't mention before is that it also paves the way for the statements to be sent in batches rather than one-by-one, which the xAPI seems to support, and I think that would be a very good thing.

ryasmi:
I think the interval for the cron needs to be configurable. @garemoko some people may want this data in real time rather than every five minutes. Unfortunately this is just a limitation of PHP, maybe in the future we can find a way to write this plugin in Node so that we can have data sent in real-time and non-blocking.

micaherne:
From Moodle 2.7 onwards the frequency of the scheduled tasks is configurable through the admin interface, and they can run in parallel, so you could send them every minute or quicker if you had your cron job set up to run frequently enough.

garemoko:
Ok, that makes sense as a setting then; thanks for explaining.

jlowe64:
And this plugin is only compatible with 2.8 and higher. It would be possible to have "real time" data if you set the cron to be every minute. We were thinking about doing this in the beginning, but felt it was best just to see how the logstore system handles things. I'm guessing, though, if you had thousands of people doing stuff on your site at once, it would probably have some fair performance issues. None of our users are that high use, so we haven't been able to test in that way.

This is a reasonable PR and definitely worth a think. I'll take a look at what you've done.

jlowe64:
@micaherne I've tested this by using the same vendor settings as 0.3.0 and when in background mode it doesn't seem to be shooting the statements over. I don't see any errors in the error log or when I manually do the cron. I have set the scheduled task to run every minute, so it should have worked.

I can't really see any reason why it wouldn't work. Can you test it on your end?

It seems to run just fine when background mode works.

micaherne:
I reinstalled the vendor packages and it's still working for me. Could you check some things:
- do you have the logstore_xapi_log table in your database?
- are there any records in it?
- what happens if you run the scheduled task by going to /admin/tool/task/cli and running php schedule_task.php --execute=\logstore_xapi\task\emit_task (with double backslashes if you're on Linux)

Also, I don't quite understand what you mean by "it seems to run just fine when background mode works" - was it working OK for you then stopped?

If you can check these things we should be able to work out where it's breaking.

(Sorry for taking so long to get back to you, by the way - I haven't been at work for most of the week)

jlowe64:
Sorry I didn't make myself very clear. 

I installed your plugin fresh on my system. It went through the upgrade process and installed that table in the DB. It caught an event (viewing a course), however when it came time for the scheduled task, I never saw that entry make it to the LRS. When I switched it back to non-background mode, it shot all events over to the LRS as normal.

I just finished manually running the task and got this: 

Scheduled task: Emit records to LRS
Sent learning records to LRS.
... used 7 dbqueries
... used 0.37417721748352 seconds
Task completed.

And I was able to confirm that they were in the LRS, and they were removed from the table. I'm just gonna assume that this is due to weirdness in my dev environment!

I'll be merging this as soon as I get a couple things cleaned up in the libraries.

micaherne:
Thanks, glad it worked!

PR #39, opened 2016-01-13
Support translator returning multiple events

garemoko:
The idea here is that where one event in Moodle needs to trigger multiple xAPI statements, the $translatorevent returned by the translator will contain an array with two properties, 'multiple_events' and 'events'. 'multiple_events' will contain the value true to indicate that the assoc. array contains multiple events. 'events' will contain and array of translatorevents.

An alternative approach would be to re-wire the translator so that it always returns an array of events. That's a more significant change.

@ryansmith94 @jlowe64 any thoughts on the general approach? 

https://github.com/jlowe64/moodle-logstore_xapi/issues/32


classes/log/store.php
```diff
@@ -110,8 +110,16 @@ public function process_events(array $events) {
             }
 
             // $this->error_log_value('translatorevent', $translatorevent);
-            $xapievent = $xapicontroller->createEvent($translatorevent);
-            // $this->error_log_value('xapievent', $xapievent);
+            if (isset($translatorevent['multiple_events']) && $translatorevent['multiple_events'] == true) {
```
ryasmi: Always return an array of events instead to reduce the complexity here.

garemoko:
This now expects an array always. I also made it easier to turn the logging on and off because it was doing my head in.

classes/log/store.php
```diff
@@ -96,31 +97,33 @@ public function process_events(array $events) {
...
             $moodleevent = $moodlecontroller->createEvent($event);
             if (is_null($moodleevent)) {
                 continue;
             }
-
-            // $this->error_log_value('moodleevent', $moodleevent);
-            $translatorevent = $translatorcontroller->createEvent($moodleevent);
-            if (is_null($translatorevent)) {
+            $this->error_log_value('moodleevent', $moodleevent);
+            $translatorevents = $translatorcontroller->createEvents($moodleevent);
```
garemoko, line 107: note: createEvents instead of createEvent

Will need changing in the translator.

ryasmi:
Nice work with the logging, it wasn't something I had to use much because I develop the tests first so I probably don't spend as much time debugging.

ryasmi:
Is this ready to merge @garemoko?

classes/log/store.php
```diff
@@ -96,22 +97,22 @@ public function process_events(array $events) {
...
+            if (is_null($translatorevents)) {
```
ryasmi: My recent change to the controller in the PR for multiple translator events means that this check will no longer be necessary because it just returns an empty array rather than `null`. This can be left as it is for now though, it has no effect.

garemoko:
@ryansmith94 yes this is ready to merge if you're happy with it

garemoko:
@ryansmith94 @jlowe64 is this PR waiting for me to do anything? (I know I have actions on other ones)

ryasmi:
I think it's just waiting on a response to a [code comment](https://github.com/jlowe64/moodle-logstore_xapi/pull/39#discussion_r49883931) @garemoko. From what I can see everything else looks fine.

garemoko:
@ryansmith94 sorry I misunderstood the comment, I though you meant that the code could remain as is. I've removed that check now.

I did consider replacing it with a check for an empty array, but in that case the foreach would do nothing anyway so it would be a pointless check I think.

ryasmi:
No worries @garemoko. Yeah I agree that it would be a pointless check. Thanks for the update.

PR #44, opened 2016-01-27
Send statements to LRS in batches

micaherne:
Hi, here is a patch implementing https://github.com/jlowe64/moodle-logstore_xapi/issues/43.

It depends on these changes to library code:
- https://github.com/LearningLocker/Moodle-Log-Expander/pull/25
- https://github.com/LearningLocker/Moodle-xAPI-Translator/pull/34
- https://github.com/LearningLocker/xAPI-Recipe-Emitter/pull/27

### DIFF
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -105,8 +105,23 @@ class store extends php_obj implements log_writer {
         $this->error_log_value('moodleevent', $moodleevents);
         $translatorevents = $translatorcontroller->createEvents($moodleevents);
         $this->error_log_value('translatorevents', $translatorevents);
-        $xapievents = $xapicontroller->createEvents($translatorevents);
-        $this->error_log_value('xapievents', $xapievents);
+
+        if (empty($translatorevents)) {
+            return;
+        }
+
+        // Split statements into batches.
+        $eventbatches = array($translatorevents);
+        $maxbatchsize = get_config('logstore_xapi', 'maxbatchsize');
+
+        if (!empty($maxbatchsize) && $maxbatchsize < count($translatorevents)) {
+            $eventbatches = array_chunk($translatorevents, $maxbatchsize);
+        }
+
+        foreach ($eventbatches as $translatoreventsbatch) {
+            $xapievents = $xapicontroller->createEvents($translatoreventsbatch);
+            $this->error_log_value('xapievents', $xapievents);
+        }
 
     }
 
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -95,22 +95,19 @@ class store extends php_obj implements log_writer {
         $translatorcontroller = new translator_controller();
 
         // Emits events to other APIs.
-        foreach ($events as $event) {
-            $event = (array) $event;
-            $this->error_log('');
-            $this->error_log_value('event', $event);
-            $moodleevent = $moodlecontroller->createEvent($event);
-            if (is_null($moodleevent)) {
-                continue;
-            }
-            $this->error_log_value('moodleevent', $moodleevent);
-            $translatorevents = $translatorcontroller->createEvents($moodleevent);
-            $this->error_log_value('translatorevents', $translatorevents);
-            foreach ($translatorevents as $index => $translatorevent) {
-                $xapievent = $xapicontroller->createEvent($translatorevent);
-                $this->error_log_value('xapievent', $xapievent);
-            }
+        foreach ($events as $index => $event) {
+            $events[$index] = (array) $event;
         }
+
+        $this->error_log('');
+        $this->error_log_value('events', $events);
+        $moodleevents = $moodlecontroller->createEvents($events);
+        $this->error_log_value('moodleevent', $moodleevents);
+        $translatorevents = $translatorcontroller->createEvents($moodleevents);
+        $this->error_log_value('translatorevents', $translatorevents);
+        $xapievents = $xapicontroller->createEvents($translatorevents);
+        $this->error_log_value('xapievents', $xapievents);
+
     }
 
     private function error_log_value($key, $value) {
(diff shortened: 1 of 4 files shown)


classes/log/store.php
```diff
@@ -95,22 +95,34 @@ public function process_events(array $events) {
...
+        $moodleevents = $moodlecontroller->createEvents($events);
+        $this->error_log_value('moodleevent', $moodleevents);
+        $translatorevents = $translatorcontroller->createEvents($moodleevents);
+        $this->error_log_value('translatorevents', $translatorevents);
+
+        if (empty($translatorevents)) {
+            return;
+        }
+
+        // Split statements into batches.
```
ryasmi, line 113: Good idea.

settings.php
```diff
@@ -41,4 +41,8 @@
     $settings->add(new admin_setting_configcheckbox('logstore_xapi/backgroundmode',
         get_string('backgroundmode', 'logstore_xapi'),
         get_string('backgroundmode_desc', 'logstore_xapi'), 0));
+
+    $settings->add(new admin_setting_configtext('logstore_xapi/maxbatchsize',
```
ryasmi, line 45: Nice, although 500 may be too high as a default.

ryasmi:
Good work, however, the tests need to pass and I think we should discuss the [default batch size](https://github.com/jlowe64/moodle-logstore_xapi/pull/44/files#r50970624) as 500 may be too high. The tests are probably failing because the dependencies need updating.

settings.php
jlowe64, line 45: Yeah, I agree with @ryansmith94. I think it should be less than 500 as default. Our largest customer only does about 4,000 statements in a day, and I suspect many more would be a lot lower than that.

ryasmi, line 45: I think around 20-30 would actually be fine as a default in most cases.

jlowe64, line 45: That would be awesome actually. :+1:

ryasmi:
Regarding the default batch size, @jlowe64 and I seem to think it should be changed to around 20-30 @micaherne.

settings.php
garemoko, line 45: Point of interest: Watershed requests with a limit of 500 statements when pulling statements from another LRS. I'd actually be happy with a default batch size of unlimited and then people can introduce a batch size if required.

classes/log/store.php
```diff
@@ -95,22 +95,34 @@ public function process_events(array $events) {
...
+        $eventbatches = array($translatorevents);
+        $maxbatchsize = get_config('logstore_xapi', 'maxbatchsize');
+
+        if (!empty($maxbatchsize) && $maxbatchsize < count($translatorevents)) {
+            $eventbatches = array_chunk($translatorevents, $maxbatchsize);
+        }
```
garemoko, line 119: +1 to batching between translator and emitter. This will ensure that the batch size is respected even when 1 Moodle event gets turned into several translator events.

settings.php
ryasmi, line 45: My thinking here @garemoko, is that on a load balanced LRS, sending lots of requests with a reasonable amount of data will actually be quicker than sending a huge number of statements.

garemoko:
The tests are certainly failing because dependancies :-) presumably there will be other PRs to update those.

settings.php
ryasmi, line 45: We should consider that it may be more than just Moodle pushing statements to the LRS. For example, imagine the LRS is not load balanced, if Moodle sends 500 statements just before another service X sends 1 statement, X may have to wait a long time while Moodle hogs the LRS. On the other hand, if Moodle sends 20 statements, X will have far less time to wait. The average plugin user that just dumps this plugin on their Moodle installation without changing the settings probably won't consider this situation and they shouldn't have to.

classes/log/store.php
ryasmi, line 119: Yeah +1 for this from me too :smile_cat:

ryasmi:
Yeah I think that's probably best @garemoko.

jlowe64:
@ryansmith94 I fixed the test previously, so I merged my changes just now.

ryasmi:
@jlowe64, that seems to have broken the master branch, which shouldn't really ever break. Use a separate branch and a PR next time :+1: In the meantime, the Travis build needs to be inspected and fixed.

jlowe64:
@ryansmith94 I fixed it.

ryasmi:
Nice one cheers @jlowe64. I think the reason for the fail is that the PR in the Translator hasn't been merged yet. I'm waiting on a change to the default batch size there.

jlowe64:
Yup, I was just gonna say that. I think the updates to the libraries need to be done first, then we can handle this end after that.

micaherne:
Thanks for the feedback, everyone. I've changed the default batch size to 30 as suggested. Hopefully it'll all work once the dependencies are sorted out, but if anything needs changed in the patch I can look at it tomorrow.

ryasmi:
Awesome, thanks for your time and effort @micaherne :+1:

ryasmi:
@jlowe64, I think for this to pass we need to:
1. Release a new 1.0.0 version of the components.
2. Update the dependencies using Composer.

jlowe64:
@ryansmith94 Are you saying to skip 0.6.0 and just go to 1.0.0 of the libraries?

ryasmi:
Yeah because they [broke backwards compatibility](http://semver.org/).

jlowe64:
@micaherne I'll be doing an official release of this plugin tomorrow. I've done the libraries today.

micaherne:
Awesome, thanks @jlowe64!

jlowe64:
@micaherne so, I think you'll have to update your PR to use the new libraries that were released. The master of this repo now holds the correct composer files. Master is currently failing, but that's only because your changes haven't been implemented yet. So if you could update to use the newest libraries, we can rush this back to a passing state!

ryasmi:
@jlowe64, the build just needed restarting mate. Travis should ideally restart the build automatically when new commits are made to the branch your merging into, but never mind. Should be good now :+1:

jlowe64:
AHHHHHH, I'm dumb. Thank you.

ryasmi:
Haha no worries.

micaherne:
Do I still need to do something? I'm not completely following this!

jlowe64:
@micaherne you don't have to do anything! I just didn't know travis had a little restart button. :)

PR #45, opened 2016-01-27
Test fix

jlowe64:


jlowe64:
Libraries need fixing and updating before this will pass.

ryasmi:
Ah the actual reason for the issue here is that we need to make releases of the other repos and update the dependencies.

ryasmi:
Yeah what you said haha. That timing

jlowe64:
Dupe work.

Issue #54, opened 2016-03-24
Event filter


CinecaElearning:
**Motive**
A lot of events are now sent to the LRS; Moodle admins could decide which are interesting and which can be discarded.

**Result**
A menu of the events that can be sent in the plugin settings.

**Additional information**
A possible implementation is developed in our branch xapi_store_2016031401. Events that are available to be sent are taken from the translator routes and presented in the settings.php. A setting for the guest generated events is present too. When the event is to be sent, the __construct parses the setting and the is_event_ignored decides if the event is to be sent or not.

ryasmi:
Hey if you have a branch that does this already feel free to make a PR and we'll take a look. I don't know if @jlowe64 or @garemoko have any input on this.

CinecaElearning:
Thanks! I had the opportunity to talk about this via email with @jlowe64, which kindly explained to me how to propose our code (I'm fairly new to github). I'll make a PR but first I want to open other issues that our code already covers. I hadn't had the time to open them yesterday.

jlowe64:
@CinecaElearning I would suggest making a PR so we can see the code. 

I've had, I think, one other group request this sort of thing and I think some folks may be into the idea of filtering just interesting stuff to the LRS. For my organization specifically, we actually prefer having more information rather than less, and the number of events being sent over has been mitigated by batching.

I think if we integrate it, the default setting should be that _all_ events are passed through, and that only in the case of a filter should it start filtering. We should also be mindful of performance issues with checks, as specifically for people with large numbers of users, performance is a priority.

CinecaElearning:
@jlowe64 I created the PR. If I recall correctly, by now, the code needs the admin to specify the events that are to be sent. Changing the default to send all is not a problem though.

garemoko:
Yeh +1 to this as a feature. I'd want all events on by default though.

timothyjoh:
+1 to this, would make it much better for us

davidpesce:
@CinecaElearning going through old issues and attempting to get the status of them. Any chance you were able to create this PR? If not, would you be able to forward the email from @jlowe64 explaining what needs to be done in the PR to davidpesce@gmail.com?

davidpesce:
This is complete as of v1.5.0

PR #55, opened 2016-03-25
New completed_course event, event filter settings, mbox based user identification, event dispatch error detection

CinecaElearning:


db/tasks.php
```diff
@@ -28,10 +28,10 @@
     array(
         'classname' => '\logstore_xapi\task\emit_task',
         'blocking' => 0,
-        'minute' => '*/5',
+        'minute' => '*/1',
```
ryasmi, line 31: Pretty sure this shouldn't be changed, I think this was discussed previously.

tests/CourseCompletedTest.php
```diff
@@ -0,0 +1,3 @@
+<?php namespace Tests;
+
+class CourseCompletedTest extends TestCase {}
```
ryasmi, line 3: Why is this here?

classes/log/store.php
```diff
@@ -53,12 +54,21 @@ class store extends php_obj implements log_writer {
 
     protected $loggingenabled = false;
 
+    /** @var bool $logguests true if logging guest access */
+    protected $logguests;
+
+    /** @var array $routes An array of routes to include */
+    protected $routes = array();
```
ryasmi: Use `[]` instead of `array()`.

classes/log/store.php
```diff
@@ -53,12 +54,21 @@ class store extends php_obj implements log_writer {
...
+
     /**
      * Constructs a new store.
      * @param log_manager $manager
      */
     public function __construct(log_manager $manager) {
         $this->helper_setup($manager);
+        $this->logguests = $this->get_config('logguests', 1);
+        $routes = $this->get_config('routes', '');
+        $this->routes = $routes === '' ? array() : explode(',', $routes);
```
ryasmi: Use `[]` instead of `array()`.

classes/log/store.php
```diff
@@ -68,6 +78,17 @@ public function __construct(log_manager $manager) {
      *
      */
     protected function is_event_ignored(event_base $event) {
+
```
ryasmi: Any thoughts on this @jlowe64?

classes/log/store.php
```diff
@@ -36,6 +36,7 @@
 use \XREmitter\Controller as xapi_controller;
 use \XREmitter\Repository as xapi_repository;
 use \MXTranslator\Controller as translator_controller;
+use \MXTranslator\Events\Event as Event;
```
ryasmi, line 39: Is this used?

ryasmi:
What happens if a new event is added here? Is it switched on or off by default when a user upgrades?

classes/log/store.php
```diff
@@ -53,12 +54,21 @@ class store extends php_obj implements log_writer {
...
+    /** @var array $routes An array of routes to include */
+    protected $routes = array();
+
     /**
      * Constructs a new store.
      * @param log_manager $manager
      */
     public function __construct(log_manager $manager) {
         $this->helper_setup($manager);
+        $this->logguests = $this->get_config('logguests', 1);
```
ryasmi, line 69: Assuming `1` is a default, it should probably be changed to `true` right?

classes/log/store.php
```diff
@@ -68,6 +78,17 @@ public function __construct(log_manager $manager) {
      *
      */
     protected function is_event_ignored(event_base $event) {
+
+        if ((!CLI_SCRIPT or PHPUNIT_TEST) and !$this->logguests) {
```
ryasmi: Use `&&` instead of `and` and use `||` instead of `or`. The two have different precedence and I don't want other contributors to get tripped up by that.

classes/log/store.php
```diff
@@ -68,6 +78,17 @@ public function __construct(log_manager $manager) {
...
+            // Always log inside CLI scripts because we do not login there.
+            if (!isloggedin() or isguestuser()) {
+                return true;
+            }
+        }
+        if (!in_array($event->eventname, $this->routes))
```
ryasmi: Wrap the contents of the `if` in a block `{ ... }` for consistency.

classes/log/store.php
```diff
@@ -68,6 +78,17 @@ public function __construct(log_manager $manager) {
      *
      */
     protected function is_event_ignored(event_base $event) {
+
+        if ((!CLI_SCRIPT or PHPUNIT_TEST) and !$this->logguests) {
+            // Always log inside CLI scripts because we do not login there.
+            if (!isloggedin() or isguestuser()) {
```
ryasmi: Condition can be moved into the previous `if` condition.

db/tasks.php
CinecaElearning, line 31: sure, it was just added in my code because I can't wait 5 minutes to debug every time. Could it be an interesting improvement moving that in the settings?

classes/log/store.php
```diff
@@ -118,11 +146,18 @@ public function process_events(array $events) {
             $eventbatches = array_chunk($translatorevents, $maxbatchsize);
         }
 
+		$translator_event = new Event();
+		$translator_event_read_return = @$translator_event->read(array());
```
ryasmi: This looks a bit odd, what's going on here?

classes/task/emit_task.php
```diff
@@ -50,10 +50,13 @@ public function execute() {
         $store = new store($manager);
 
         $events = $DB->get_records('logstore_xapi_log');
-        $store->process_events($events);
-
-        $DB->delete_records_list('logstore_xapi_log', 'id', array_keys($events));
-
+        $store_return = $store->process_events($events);
+		foreach(array_keys($store_return) as $event_id)
+			{
```
ryasmi: Opening brace needs to be on the same line as the `foreach` according to the [style guide](http://jlowe64.github.io/moodle-logstore_xapi/).

classes/task/emit_task.php
```diff
@@ -50,10 +50,13 @@ public function execute() {
...
+			if ($store_return[$event_id] == 'success')
```
ryasmi: I don't think this will work as expected, a Moodle event can produce multiple xAPI events.

ryasmi:
I think this needs splitting into multiple PRs, one that addresses each issue because this definitely seems to address more than #54.

classes/log/store.php
CinecaElearning: No problem with that, but I copied from the database store which is Moodle core. Should I go on?

CinecaElearning:
Sure, I was saying the same thing yesterday: "I'll make a PR but first I want to open other issues that our code already covers". This code is the sum of different things we needed and now we are sharing it. It's a bit confusing, I know.

classes/task/emit_task.php
CinecaElearning: Thanks, I'll see to that.

db/tasks.php
ryasmi, line 31: Yeah certainly could.

classes/log/store.php
ryasmi: Yeah please continue with the change, our style is likely to be different to the Moodle core.

classes/log/store.php
CinecaElearning, line 69: This too is copied from the Moodle core (database store) which already implements the guests' events filtering.

classes/log/store.php
CinecaElearning, line 39: Yes, in line 149, for the odd thing:
$translator_event = new Event();

classes/log/store.php
ryasmi, line 69: @jlowe64 will be better placed to comment on this.

classes/log/store.php
CinecaElearning: Done

classes/log/store.php
CinecaElearning: Done

classes/task/emit_task.php
CinecaElearning: Done

classes/task/emit_task.php
CinecaElearning: I made some tests with simple events, especially the course completed and everything went fine. Can you show me some Moodle events that produce multiple xAPI events so I can test them?

ryasmi: `\mod_quiz\event\attempt_reviewed` as shown in the [Translator Controller](https://github.com/LearningLocker/Moodle-xAPI-Translator/blob/master/src/Controller.php#L33)

db/tasks.php
CinecaElearning, line 31: I saw now that you can customize it here:
http://yourmoodle.com/admin/tool/task/scheduledtasks.php?action=edit&task=logstore_xapi%5Ctask%5Cemit_task

not sure if it will override the db/tasks.php setting, but I assume so. I'll give it a try.

tests/CourseCompletedTest.php
CinecaElearning, line 3: We implemented the course completed event too. I assumed I needed to put this file too along with the other test files that correspond to the other events. Shall I remove this file?

classes/log/store.php
CinecaElearning: To catch the dispatch status and assign it to the correct event id, I had to parse the returned structure (see under). The event id is inside an associative array which key is "http://lrs.learninglocker.net/define/extensions/moodle_logstore_standard_log". Instead of hardcoding the key here too, I read it from the Event, faking a read to get the context_ext_key which contains this key. Since I was afraid the key could change, invalidating the structure parsing, I decide to go along this path.

ryasmi: Where is `$xapievents['last_action_result']` set?

CinecaElearning:
About "What happens if a new event is added here? Is it switched on or off by default when a user upgrades?", it does nothing at this time. If the default is to send everything, this will lead to an event that is not sent, which is a problem. By now, the list of the events that are to be sent, is imploded separated by a comma in the config table in the DB. An upgrade procedure should take into account that if new events are added, they have to be added in the corrisponding config DB line too. We hadn't that in mind because we mentally followed the database store configuration, which by default doesn't send any event.

classes/log/store.php
CinecaElearning: in the emitter repository (for which I believe I hadn't had the time to make a PR yet).

classes/log/store.php
CinecaElearning: Done

classes/log/store.php
CinecaElearning: Done

classes/task/emit_task.php
```diff
@@ -51,12 +51,12 @@ public function execute() {
...
         $events = $DB->get_records('logstore_xapi_log');
         $store_return = $store->process_events($events);
-		foreach(array_keys($store_return) as $event_id)
-			{
-			if ($store_return[$event_id] == 'success')
+		foreach(array_keys($store_return) as $event_id) {
+			if ($store_return[$event_id] == 'success') {
 				$DB->delete_records_list('logstore_xapi_log', 'id', array($event_id));
-			else {}	// ERROR TRASMITTING THE EVENT, DO NOTHING
+			} else {
```
ryasmi, line 57: Probably just remove this else.

classes/log/store.php
```diff
@@ -147,14 +146,16 @@ public function process_events(array $events) {
...
-		$translator_event_read_return = @$translator_event->read(array());
+		$translator_event_read_return = @$translator_event->read([]);
 
-		$sent_events = array();
+		$sent_events = [];
         foreach ($eventbatches as $translatoreventsbatch) {
             $xapievents = $xapicontroller->createEvents($translatoreventsbatch);
-			foreach(array_keys($xapievents) as $key)
-				if (is_numeric($key))
+			foreach(array_keys($xapievents) as $key) {
```
ryasmi, line 154: Awaiting https://github.com/LearningLocker/xAPI-Recipe-Emitter/issues/26

ryasmi:
Is it potentially a better solution to either:
- store the events that we want to be ignored; or
- store an additional option that determines if no events should be ignored?

ryasmi:
Also @CinecaElearning can you please change the title of this PR to something more meaningful so that it's searchable in the issue tracker for other users?

db/tasks.php
garemoko, line 31: This doesn't make sense to add to settings because it's a default setting that's already customizable in the standard Moodle settings. Also note that this will never run more often than the cron is configured to run so my vote is leave this as running every minute by default. 

Happy to explain how this works in more detail if helpful.

garemoko:
RE https://github.com/jlowe64/moodle-logstore_xapi/pull/55#issuecomment-201286891 - this also needs splitting out into PRs for the 3 contributing repositories. See https://github.com/jlowe64/moodle-logstore_xapi/blob/master/docs/design.md

ryasmi:
I think it already has @garemoko

db/tasks.php
jlowe64, line 31: I agree with @garemoko. It shouldn't be a setting in this plugin.

classes/log/store.php
jlowe64, line 69: @ryansmith94 It seems to be pretty standard Moodle syntax, if not ugly and against our coding standards for the rest of the plugin. I've never logged guest access, so I don't know the details. From my minor bit of research, it does seem standard, and is in the database logstore plugin exactly like this.

db/tasks.php
CinecaElearning, line 31: @garemoko I was saying the same thing in the comment above yours (to be precise, I was saying that I had to be sure about it, but now you are confirming it, so that's perfect). Note though that the default is not 1 minute. I changed it to 1 minute because the default is 5.

ryasmi, line 31: Ok thanks @garemoko @jlowe64 :+1:

CinecaElearning:
RE https://github.com/jlowe64/moodle-logstore_xapi/pull/55#issuecomment-201442635
I would go with "store the events that we want to be ignored". This if the default is to send every event. I'm not very convinced about it though. When you install the plugin, if the default is "send all", you start sending a lot of data. The default frequency is 30 event every 5 minutes. If on the Moodle there is a lot of action and the admin is not completely aware, the sending table can easily become huge.

classes/log/store.php
ryasmi, line 69: Ok thanks @jlowe64

classes/log/store.php
ryasmi: Ok no worries.

tests/CourseCompletedTest.php
ryasmi, line 3: Ah right yeah fair enough.

ryasmi, line 3: Why doesn't it have any assertions though?

CinecaElearning, line 3: It's a modified copy of the CourseViewedTest.php which is empty too.

ryasmi, line 3: Ok, we need to look into that @jlowe64

davidpesce:
Functionality from this was moved to PR #104

Issue #56, opened 2016-03-25
Course complete event


CinecaElearning:
**Motive**
the course completed event is not sent

**Result**
The event is added here, in the translator and in the emitter. Eventually the grade could be sent along and maybe the expander could be affected too.

**Additional information**
A possible implementation is developed in our branch xapi_store_2016031401. The event and the test are added in the moodle-logstore_xapi and the same has been done in the translator and in the emitter.

timothyjoh:
+1 for this, we need it

DaTebe:
+1, with grades in the complete event. For me it would be a cool thing to visualize if there is an Gaussian distribution =)

jlowe64:
Completed by:

https://github.com/xAPI-vle/moodle-logstore_xapi/pull/75
https://github.com/LearningLocker/Moodle-Log-Expander/pull/33
https://github.com/LearningLocker/Moodle-xAPI-Translator/pull/41
https://github.com/LearningLocker/xAPI-Recipe-Emitter/pull/34

PR #75, opened 2016-05-26
Test for course_completed added.

caperneoignis:
This is the test requested by @ryansmith94 for the plugin. All test passed. Let me know if I did not make the test properly. It looks correct. Let me know if there is anything I need to look at after the test pass to ensure the item is being formed properly.


caperneoignis:
This should resolve issue #56. Forgot to put that issue in the commit.

PR #85, opened 2016-08-23
Update store.php

deedey:
Added moodle config proxy parameters

### DIFF
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -154,12 +154,17 @@ class store extends php_obj implements log_writer {
      * @return xapi_repository
      */
     private function connect_xapi_repository() {
-        return new xapi_repository(new tincan_remote_lrs(
+        global $CFG;
+        $remote_lrs = new tincan_remote_lrs(
             $this->get_config('endpoint', ''),
             '1.0.1',
             $this->get_config('username', ''),
             $this->get_config('password', '')
-        ));
+        );
+        if (!empty($CFG->proxyhost)) {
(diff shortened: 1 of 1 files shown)


ryasmi:
Ready to merge for the next minor release @jlowe64 👍

ryasmi:
Actually @jlowe64, before merging this, ensure that the tincan php library has been updated in the composer.json file and composer.lock file.

davidpesce:
@ryansmith94 do you remember which version of TinCanPHP you were referencing above?

ryasmi:
No I don't sorry @davidpesce. What's the reasoning for using submodules? My only qualm would be the bad press I've read about using submodules in the *past*.

garemoko:
I would much rather we use submodules than composer. I'm a big submodules fan.

davidpesce:
I think it did get some bad press for being used wrong. I'm fine either way but do prefer submodules. I don't want to rock the boat so I'll leave it as is!

davidpesce:
TinCan PHP 1.0 is a required dependency in this PR: https://github.com/xAPI-vle/moodle-logstore_xapi/pull/95

Once that has been merged, this should be OK to merge as well.

davidpesce:
@ryansmith94 should I create another issue/PR with the camelcase fixes? Or should I ignore it?

ryasmi:
Nah ignore that for now, need to just have a sweep through the repository and make code climate happy.

PR #94, opened 2017-03-02
Update moodle-logstore_xapi to use php 5.5 to 7.0

blm-unikent:
The composer.json changes should allow Moodle servers with php 5.5, 5.6 and 7.0 to run this plugin.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: php
-php: 5.5
+php:
+    - 5.5
+    - 5.6
+    - 7.0
 install: composer install --no-interaction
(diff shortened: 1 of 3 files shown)


ryasmi:
Hi @brianlmerritt-kent, could you please also change `5.5` to `7.0` as the [PHP version in the Travis config](https://github.com/xAPI-vle/moodle-logstore_xapi/blob/master/.travis.yml#L2). I don't think that won't fix the tests, it looks like the version format in your composer.json file is incorrect.

blm-unikent:
Done as different pull request - delete this one

PR #95, opened 2017-03-02
Brianlmerritt kent use php 5.6 or 7

blm-unikent:
The composer.json changes should allow Moodle servers with php 5.5, 5.6 and 7.0 to run this plugin. Fix version issue in composer.json and added travis tests

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     },
     "require": {
         "php": "~5.5 || ^7.0.3",
-        "rusticisoftware/tincan": "1.0.0 as 0.12.0",
+        "rusticisoftware/tincan": "1.0.0",
         "learninglocker/moodle-log-expander": "~1.1",
         "learninglocker/moodle-xapi-translator": "~1.1",
         "learninglocker/xapi-recipe-emitter": "~1.1"
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
         "learninglocker/statementfactory": "~0.0"
     },
     "require": {
-        "rusticisoftware/tincan": "0.40.0",
+        "rusticisoftware/tincan": "~1.0 as 0.12.0",
         "learninglocker/moodle-log-expander": "~1.1",
         "learninglocker/moodle-xapi-translator": "~1.1",
         "learninglocker/xapi-recipe-emitter": "~1.1"
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,8 @@
         "learninglocker/statementfactory": "~0.0"
     },
     "require": {
-        "rusticisoftware/tincan": "~1.0 as 0.12.0",
+        "php": "~5.5 || ^7.0.3",
+        "rusticisoftware/tincan": "1.0.0 as 0.12.0",
         "learninglocker/moodle-log-expander": "~1.1",
         "learninglocker/moodle-xapi-translator": "~1.1",
         "learninglocker/xapi-recipe-emitter": "~1.1"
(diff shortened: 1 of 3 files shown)


ryasmi:
Hi @brianlmerritt-kent, thank you for your contribution, I'm pretty happy with this, but could you please link me to some documentation for the version syntax you're using in `1.0.0 as 0.12.0`? I'm not sure what the exact behaviour is of the `as` in there.

ryasmi:
Never mind I've found it https://getcomposer.org/doc/articles/aliases.md#require-inline-alias

ryasmi:
What's the reason for your use of the `as`? I'm assuming it's so that the LearningLocker dependencies work, however, we also control the source of those dependencies so we should just update them to use 1.0.0.

blm-unikent:
??Some learninglocker dependencies are asking for 0.12.0, so this avoids necessity of accepting less than stable



________________________________

ryasmi:
The Learning Locker dependencies

- [Moodle Log Expander](https://github.com/LearningLocker/Moodle-Log-Expander)
- [Moodle to xAPI Translator](https://github.com/LearningLocker/Moodle-xAPI-Translator)
- [xAPI Recipe Emitter](https://github.com/LearningLocker/xAPI-Recipe-Emitter)

blm-unikent:
Exactly!


https://github.com/LearningLocker/xAPI-Recipe-Emitter/blob/master/composer.json?



________________________________

ryasmi:
Ok so I think we can apply the same changes from here in the Learning Locker dependencies.

- [composer.json in the Emitter](https://github.com/LearningLocker/xAPI-Recipe-Emitter/blob/master/composer.json#L15)
- [travis.yml in the Emitter](https://github.com/LearningLocker/xAPI-Recipe-Emitter/blob/master/.travis.yml#L2)
- [travis.yml in the Translator](https://github.com/LearningLocker/Moodle-xAPI-Translator/blob/master/.travis.yml#L2)
- [travis.yml in the Expander](https://github.com/LearningLocker/Moodle-Log-Expander/blob/master/.travis.yml#L2)

Obviously the composer.lock file will also need to change in the Emitter.

ryasmi:
Once those changes have been made we can just use 1.0.0 and drop the use of `as` which is definitely preferred.

blm-unikent:
?Yes, agreed

________________________________

davidpesce:
Added necessary changes to dependencies (Emitter, Translator, and Expander)

- Translator - https://github.com/LearningLocker/Moodle-xAPI-Translator/pull/52
- Expander - https://github.com/LearningLocker/Moodle-Log-Expander/pull/44
- Emitter - https://github.com/LearningLocker/xAPI-Recipe-Emitter/pull/37

davidpesce:
@ryansmith94 is this one approved for merge?

composer.json
```diff
@@ -12,7 +12,8 @@
         "learninglocker/statementfactory": "~0.0"
     },
     "require": {
-        "rusticisoftware/tincan": "0.40.0",
+        "php": "~5.5 || ^7.0.3",
+        "rusticisoftware/tincan": "1.0.0 as 0.12.0",
```
ryasmi: Can we change this to just be 1.0.0 @davidpesce?

ryasmi (changes requested):

composer.json
davidpesce: Should be all set!

ryasmi (approved):

PR #105, opened 2017-06-05
V1.5.0

davidpesce:
v1.5.0 release


davidpesce:
@ryansmith94 can you take a look at this one. When I get the tests working in translator, they fail in logstore. When I get it working in logstore, they fail in translator. Any ideas on what is causing this?

PR #108, opened 2017-06-07
course_module_completed_event

nikidev:
Course module completed event   implemented   similarly to Course Completed event  using  
'\core\event\course_module_completion_updated'


ryasmi:
Just rejected this PR since the PR itself obviously shouldn't be merged here. I'm sure the code itself is fine. Just don't want this getting merged accidentally.

davidpesce:
@nikidev this is a bit too much to skim over and provide feedback. Are you able to only submit the files you changed?

nikidev:
Sorry @davidpesce ! I thought  it would be easier  to compare  if I push the other files  together with the changes. But anyway  I was wrong. That's why I made the change. You can see it now ! :)

davidpesce:
@nikidev No worries. It's a bit to wade through, but from what I'm seeing, your events aren't implementing anything different than the base event itself. If you only extend the base event class, you won't get any additional information. 

Another thing that will help is to implement the unit tests for each of the sub-repositories (Expander, Translator, Emitter). This will help you test the code without needing to post to a Moodle instance.

Just to be clear, are you trying to report when a topic is complete? Or a specific activity?

nikidev:
I am trying to report for any module (activity) if it is completed.  For example  I have **"custom URL activity"**.   On the screenshot below  I get the event name  for the '\core\event\course_module_completion_updated'  event  in the statement for the certain activity.
I know that this is the proper statement  because  I checked it in the moodle standard log. So the emiter  works  fine and emit the statement.

eventname

But the issue is here  where  instead of name of the activity "custom url activity"  learning locker displays the name of the course (mvc 14).  

issue




So I am wondering  where I have to change the structure  of the course_module_completed_event statement  to make learning locker  to display the name of the activity.  As you know  learning locker  is taking that info from the statement  and there is a $object variable  which accept that value.

davidpesce:
Ahh, gotcha. I don't know the answer off the top of my head, but can probably get you pointed in the right direction. You'll have to update the Expander functionality to do this. It looks like you are currently only extending the base Event class functionality. I think you'll need to add something like this: 

```
'module' => $this->repo->readModule($scormid, 'scorm'),
```

in your read function.

Take a look at the ScormSubmitted.php file (within Expander repo) for additional info. It may get you where you need to go.

Also, I can see this being a good functionality to include in the plugin. Would you mind submitting a PR with the final functionality?

nikidev:
Hi @davidpesce ! 
I tried  this in the expander.

`'module' => $this->repo->readObject($opts['objectid'], $opts['objecttable'])`

But nevermind  !

I think that  the problem is  in the  emitter ! 

code




I am using  readCourse function for  'object'. I think it has to be readModule.  But readModule expects other  parameters.  Hmm.  The functionality which I am doing is working  I just want to change the  object name of the course  with that of the module. 

I think that   the  name  of the (course) in that screenshot  is taken  exactly from  object:   part of the statement.
26881853-5b14630c-4ba1-11e7-8e75-5407bb8437a5

So if  I change  that object  to receive  the module  it will emit the module name not   the  course name

nikidev:
'module' => $this->repo->readModule($scormid, 'scorm'), 

By the way this readModule  function  won't  help  because  it expects "type of module plugin (assign,quiz)" as a  second parameter.

davidpesce:
Yeah, but the only issue will be that you'll need to figure out what the module type is to make the request.

nikidev:
Here is the problem !  I don't  need to define the type.  I make it for every  module not for a specific type (assign,scorm  etc.)

davidpesce:
Yeah, but the readModule function requires the type.

nikidev:
Yeah I know it !  So, what type should I provide ?

nikidev:
Is there a way  to discuss that deeply  somewhere  on a chat  or  a call ?

davidpesce:
I'm on gitter in the LearningLocker room. https://gitter.im/LearningLocker/learninglocker

nikidev:
Alright !  I wrote you there.  Write me  when you are available.

davidpesce:
Closing. Replaced this PR with one that only has necessary files changed. Provided by original contributor. https://github.com/xAPI-vle/moodle-logstore_xapi/pull/120

PR #111, opened 2017-06-12
In background mode, unsupported events are no longer (pointlessly) stored

mikemcgowan:
When in 'background mode', *all* incoming events were being stored in `logstore_xapi_log` (for later processing by a cron task to send them to the configured LRS). However, only a relatively small subset of Moodle events is actually supported.

Hence, many of the events supposedly temporarily stored in `logstore_xapi_log` would actually be written to that table permanently - with no way of removing them - thereby wasting storage space and increasing cron processing time. (In particular, considering that `emit_task::execute` does `select * from logstore_xapi_log` which is terribly inefficient ...)

I have added a couple of lines to `store::insert_event_entries` to filter out unsupported events (before saving the supported ones to `logstore_xapi_log`).

I've also fixed three Moodle coding standards violations in the same file `store.php`, which were just whitespace-related.

### DIFF
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -100,7 +100,12 @@ class store extends php_obj implements log_writer {
 
         // If in background mode, just save them in the database
         if (get_config('logstore_xapi', 'backgroundmode')) {
-            $DB->insert_records('logstore_xapi_log', $events);
+            $events = array_filter($events, function ($event) {
+                return array_key_exists($event['eventname'], moodle_controller::$routes);
+            });
+            if (!empty($events)) {
+                $DB->insert_records('logstore_xapi_log', $events);
+            }
         } else {
             $this->process_events($events);
         }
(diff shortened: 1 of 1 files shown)


garemoko:
Sounds like a good change. 

One alternative you might want to consider while in this space is ditching the logstore_xapi_log table entirely and just using the standard log, which logstore_xapi_log is a clone of. This plugin could just keep a record of which entries in the standard log had been processed (or ignored), or even just the date up to which they had been processed.

mikemcgowan:
> you might want to consider ditching the logstore_xapi_log table entirely and just using the standard log

Hmm, not sure about that one; it's possible to disable individual logstore plugins, including the standard log! Even if it weren't possible to disable the standard log, it's possible (but admittedly unlikely) that standard log entries could be purged before the xAPI logstore had chance to send them.

I agree the repetition (of log data) feels wrong, but the only alternative is to have one logstore (the xAPI) dependent on another (the standard), which doesn't feel right either ...

mikemcgowan:
However, I'm not really sure this change is necessary afterall, due to the following code in the store's `is_event_ignored` method, which was added in b2407942b36fffba2f1e2ec7b0325c0b9fc927de

```php
        if (!in_array($event->eventname, $this->routes)) {
            // Ignore event if the store settings do not want to store it.
            return true;
        }
```

So, I'm going to close this PR.

PR #141, opened 2017-10-05
Bind Moodle $CFG->debug to $loggingenabled

nadavkav:
**Description**
Turn on xapi_logstore file debugging when Moodle debugging is turned on.
(maybe this should go into os level standard logs as well, so it not bloating inside the folder?)

**Related Issues**
none

**PR Type**
- Enhancement

### DIFF
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -65,10 +65,16 @@ class store extends php_obj implements log_writer {
      * @param log_manager $manager
      */
     public function __construct(log_manager $manager) {
+        global $CFG;
+        
         $this->helper_setup($manager);
         $this->logguests = $this->get_config('logguests', 1);
         $routes = $this->get_config('routes', '');
         $this->routes = $routes === '' ? [] : explode(',', $routes);
+        
+        if (!empty($CFG->debug) and $CFG->debug >= DEBUG_DEVELOPER) {
+            $this->loggingenabled = true;
+        }
     }
 
     /**
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -66,12 +66,10 @@ class store extends php_obj implements log_writer {
      */
     public function __construct(log_manager $manager) {
         global $CFG;
- 
         $this->helper_setup($manager);
         $this->logguests = $this->get_config('logguests', 1);
         $routes = $this->get_config('routes', '');
         $this->routes = $routes === '' ? [] : explode(',', $routes);
-
         if (!empty($CFG->debug) and $CFG->debug >= DEBUG_DEVELOPER) {
             $this->loggingenabled = true;
         }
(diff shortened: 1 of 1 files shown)


classes/log/store.php
```diff
@@ -65,10 +65,16 @@ class store extends php_obj implements log_writer {
      * @param log_manager $manager
      */
     public function __construct(log_manager $manager) {
+        global $CFG;
+
```
Nicosoft: Remove all white spaces (Moodle coding style)

classes/log/store.php
```diff
@@ -65,10 +65,16 @@ class store extends php_obj implements log_writer {
      * @param log_manager $manager
      */
     public function __construct(log_manager $manager) {
+        global $CFG;
+        
         $this->helper_setup($manager);
         $this->logguests = $this->get_config('logguests', 1);
         $routes = $this->get_config('routes', '');
         $this->routes = $routes === '' ? [] : explode(',', $routes);
+
```
Nicosoft: Remove all white spaces (Moodle coding style)

Nicosoft (changes requested):
Thank you @nadavkav 
You just need to remove all spaces in line 69 and 74 because of Moodle coding style.

nadavkav:
I removed the lines completely.

classes/log/store.php
```diff
@@ -65,10 +65,14 @@ class store extends php_obj implements log_writer {
      * @param log_manager $manager
      */
     public function __construct(log_manager $manager) {
+        global $CFG;
         $this->helper_setup($manager);
         $this->logguests = $this->get_config('logguests', 1);
         $routes = $this->get_config('routes', '');
         $this->routes = $routes === '' ? [] : explode(',', $routes);
+        if (!empty($CFG->debug) and $CFG->debug >= DEBUG_DEVELOPER) {
```
ryasmi, line 73: I might have used `&&` instead of `and` to protect against future changes, since some people don't understand there's a slight difference in precedence between the two. Not too worried about that though, just something to be aware of 😄

ryasmi (approved):

ryasmi:
Thanks for your suggestion here @Nicosoft 👍

classes/log/store.php
nadavkav: I Agree, just used an old piece of code from another plugin I have.
Better fix it, at some point.

nadavkav:
BTW, was also thinking the logs should go to the standard php os level logs. here:
https://github.com/xAPI-vle/moodle-logstore_xapi/blob/master/classes/log/store.php#L17
Maybe use "error_log($message."\r\n", 1);"
And prefix $message with something like "logstore:", 
so it can be filtered out when browsing the logs.
What do you think?

ryasmi:
Hey @nadavkav, yeah that would be awesome 👍

Issue #121, opened 2017-06-14
[Bug] enrolment_created event not working


davidpesce:
v2.0.0
Reported by @nikidev:

> The enrolment_created event seems to be not working. I made a comparison in the code between the old version and version 2.0.0 and it's the same. I double checked in the settings if the event is enabled but it is enabled by default. So this is not the problem.

Dasuntha:
Error is is  'Agent` must have exactly one identifier not `0` in `statement.context.instructor. We need to fix this in ' lib/emitter/src/Events/EnrolmentCreated.php'.

nikidev:
Great job @Dasuntha   !  :)

ryasmi:
Wow yeah good work @Dasuntha 👍

nikidev:
By the way @Dasuntha,  in the same context   do you have an idea  for that  issue: https://github.com/xAPI-vle/moodle-logstore_xapi/issues/129

 The Quiz completed  event is not emitting by the emitter.    I guess the problem is in /lib/emitter/src/Events/AttemptCompleted.php  

But I don't know  where exactly  it is. 
Everything  seems to be fine.  :)

davidpesce:
Looks like this was fixed in v2.2.2 with PR #152

PR #152, opened 2017-10-13
Fixed enrolment_created event not working

Dasuntha:
**Description**
- 

**Related Issues**
- #121

**PR Type**
- Fix

### DIFF
--- a/lib/emitter/src/Events/EnrolmentCreated.php
+++ b/lib/emitter/src/Events/EnrolmentCreated.php
@@ -37,7 +37,7 @@ class EnrolmentCreated extends Event {
             ],
             'object' => $this->read_course($opts),
             'context' => [
-            	'instructor' => $this->read_user($opts, 'instructor'),
+                'instructor' => $this->read_user($opts, 'instructor'),
             ],
         ]);
         
--- a/lib/emitter/src/Events/EnrolmentCreated.php
+++ b/lib/emitter/src/Events/EnrolmentCreated.php
@@ -37,8 +37,9 @@ class EnrolmentCreated extends Event {
             ],
             'object' => $this->read_course($opts),
             'context' => [
-                'instructor' => parent::read($opts)['actor'],
+            	'instructor' => $this->read_user($opts, 'instructor'),
             ],
         ]);
+        
     }
 }
--- a/lib/emitter/src/Events/EnrolmentCreated.php
+++ b/lib/emitter/src/Events/EnrolmentCreated.php
@@ -37,7 +37,7 @@ class EnrolmentCreated extends Event {
             ],
             'object' => $this->read_course($opts),
             'context' => [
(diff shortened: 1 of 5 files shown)


lib/emitter/src/Events/EnrolmentCreated.php
```diff
@@ -37,7 +37,7 @@ public function read(array $opts) {
             ],
             'object' => $this->read_course($opts),
             'context' => [
-                'instructor' => $this->read_user($opts, 'instructor'),
+                'instructor' => parent::read($opts)['actor'],
```
ryasmi: We should probably use `$this->read_user($opts, 'user')`, but I'm not sure this is the correct fix. I think actually the instructor needs to provided from the translator and expander. The reason issue #121 exists, is probably in the case where either:
- the user enrols themselves rather than being enrolled by a instructor, hence there's no instructor passed through
- the instructor enrols a user, but the user that's passed from the expander and translator is the instructor instead of the user being enrolled, which has also resulted in the instructor not being set

ryasmi (changes requested):

lib/emitter/src/Events/EnrolmentCreated.php
dasu2013: Hi @ryansmith94,

Thanks for your feedback. I agree with you, we need the instructor there. The problem is in EnrolmentCreated.php in translator. It didn't assign 'instructor_email' or 'user_email'. I changed it like this (added both instructor_email and user_email),
 public function read(array $opts) {
        return [array_merge(parent::read($opts)[0], [
            'recipe' => 'enrolment_created',
            'user_id' => $opts['relateduser']->id,
            'user_url' => $opts['relateduser']->url,
            'user_name' => $opts['relateduser']->fullname,
            'user_email' => $opts['relateduser']->email,
            'instructor_id' => $opts['user']->id,
            'instructor_url' => $opts['user']->url,
            'instructor_email' => $opts['user']->email,
            'instructor_name' => $opts['user']->fullname,
        ])];
    }
After I changed this and revert my previous changes in 'lib/emitter/src/Events/EnrolmentCreated.php' file, it worked properly for both self enrollments and also instructor enrollments. Thank you so much for this details reply and put me on to the correct track. I updated my pull request with latest changes. 
Dasu

ryasmi: Hey @Dasuntha, you're very welcome, thanks for your reply. I see, good work tracking that down. That's great! You're very welcome. Thanks for updating your PR, I'll get this merged and released today.

ryasmi (approved):

ryasmi:
This is in [v2.2.2](https://github.com/xAPI-vle/moodle-logstore_xapi/releases/tag/v2.2.2) now, thanks @Dasuntha :smile:

PR #112, opened 2017-06-13
It's no longer necessary to instantiate \MXTranslator\Events\Event and invoke its 'read' method just to get at the 'context' ext key

mikemcgowan:
Please see [this PR](https://github.com/LearningLocker/Moodle-xAPI-Translator/pull/66).

PR #158, opened 2018-02-22
re-did mikemcgowans fix for the context key in translator

davidpesce:
mikemcgowan submitted a PR for this fix. Was easier to recreate it from scratch.


davidpesce:
For credit, here is the original PR #112

Issue #161, opened 2018-02-23
Divide by zero error on scoreraw_submitted event


binote00:
**Description**
- if scoremin and/or scoremax are not defined when the scoreraw_submitted event is triggered then score values are not sent to the LRS. It generate a divide by zero php error.

**Version**
- branch master at commit 516bf88 version 2017061100

**Additional information**
- Tested with SCORM 1.2 basic golf scorm

davidpesce:
Thanks for submitting this! I've just merged it into master. We've got a few other upcoming changes before adding another version release.

PR #162, opened 2018-02-23
issue 161 : check if scoremax and scoremin are not 0

binote00:
I am sorry for the duplication, but here is the patch for the issue #161 

Have a nice day


davidpesce:
Thanks, @binote00! This looks good to go!

PR #166, opened 2018-03-20
added support for mod_feedback_response_submitted

VassAngels:
hey @davidpesce, @ryansmith94 said to tag you as a reviewer here.

### DIFF
--- a/src/transformer/utils/constants.php
+++ b/src/transformer/utils/constants.php
@@ -2,7 +2,7 @@
 
 namespace transformer\utils;
 
-const xapi_type = 'http://id.tincanapi.com/activitytype/lms';
+const xapi_type = 'http://id.tincanapi.com/activitytype/lms/';
 const info_extension = 'http://lrs.learninglocker.net/define/extensions/info';
 const app_type = 'http://id.tincanapi.com/activitytype/site';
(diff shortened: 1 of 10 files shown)


davidpesce:
@VassAngels thanks for submitting this! I'm hoping I can get it reviewed this week. If not, I'll be at LSCon next week and it may have to wait until the following.

davidpesce:
@VassAngels one other quick item. Can you take a look at the test that Ryan wrote? This PR will need the test associated with the event before it can be merged.

src/transformer/utils/constants.php
```diff
@@ -2,7 +2,7 @@
 
 namespace transformer\utils;
 
-const xapi_type = 'http://id.tincanapi.com/activitytype/lms';
+const xapi_type = 'http://id.tincanapi.com/activitytype/lms/';
```
garemoko, line 5: why is this added?

ryasmi: Just temporary until we create a different URL for each type.

garemoko: "there's nothing as permanent as a temporary solution"

ryasmi:
Copy, pasted, and tweaked this to work in the version 3 PR so closing this PR.

PR #175, opened 2018-06-28
Architecture

### LINKED ISSUE #174: No statements for question answers
**Description**
If I choose randomized question from the quiz options, the plugin stops working as expected when sending question submitted statements.

**Version**
- ````$plugin->version = 2018022201;````

1. add multichoice question, add true false, and add another multichoice question
2. take a quiz with the above question types. Check LRS everything will be entering fine.
3. add a random question, and use the question bank from the quiz you just created.
4. take quiz.
5. when you return to see the results you will notice errors at the bottom of the page about undefined offset.
6. go to LRS and you will notice no questions have been sent and it will say only 'page viewed' or quiz attempt started.

**Expected behaviour**
- Questions to be sent to the LRS with out issue, regardless of random question. The results should be the same with the three question types even if they are in random order.

**Actual behaviour**
- Error, nothing is sent. 

**Server information**
- MySQL 5.7, php7.0 & php 7.1 Moodle 3.5 LearningLocker.

**Client information**
- OS: Windows Server 2017, windows 10
- Browser: Chrome, firefox, IE

**Additional information**
- One line 34 of QuestionSubmitted.php the call 
```` $question = $this->expand_question(
                $opts['questions'][$questionattempt->questionid],
                $opts['questions']
            );
````

is making a call to $questionattempt->questionid however, the questionid value is pointing to an index that is not in $opts[questions], for the random question array element, only the normal questions. which causes and undefined index error.

````
stdClass::__set_state(array(
   'id' => '16',
   'questionusageid' => '8',
   'slot' => '2',
   'behaviour' => 'deferredfeedback',
   'questionid' => '5',
   'variant' => '1',
   'maxmark' => '1.0000000',
   'minfraction' => '0.0000000',
   'maxfraction' => '1.0000000',
   'flagged' => '0',
   'questionsummary' => 'adding more stuff

',
   'rightanswer' => 'True',
   'responsesummary' => 'True',
   'timemodified' => '1530118034',
   'steps' => 
  array (
    44 => 
    stdClass::__set_state(array(
       'id' => '44',
       'questionattemptid' => '16',
       'sequencenumber' => '0',
       'state' => 'todo',
       'fraction' => NULL,
       'timecreated' => '1530118007',
       'userid' => '2',
       'data' => 
      array (
      ),
    )),
    47 => 
    stdClass::__set_state(array(
       'id' => '47',
       'questionattemptid' => '16',
       'sequencenumber' => '1',
       'state' => 'complete',
       'fraction' => NULL,
       'timecreated' => '1530118017',
       'userid' => '2',
       'data' => 
      array (
        57 => 
        stdClass::__set_state(array(
           'id' => '57',
           'attemptstepid' => '47',
           'name' => 'answer',
           'value' => '1',
        )),
      ),
    )),
    50 => 
    stdClass::__set_state(array(
       'id' => '50',
       'questionattemptid' => '16',
       'sequencenumber' => '2',
       'state' => 'gradedright',
       'fraction' => '1.0000000',
       'timecreated' => '1530118034',
       'userid' => '2',
       'data' => 
      array (
        62 => 
        stdClass::__set_state(array(
           'id' => '62',
           'attemptstepid' => '50',
           'name' => '-finish',
           'value' => '1',
        )),
      ),
    )),
  ),
))
````
So as you can see in the above array, questionid is set to 5, but the actual question in the array is 3. Causing an unknown index error.

### PR DESCRIPTION
**Pull request by caperneoignis**
Working progress, making fixes for warnings I'm getting, and trying to resolve #174 will notify when this is ready for merge. but adding a pull request so others can see what I'm doing and make recommendations.

### DIFF
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -27,7 +27,7 @@ namespace logstore_xapi\log;
 defined('MOODLE_INTERNAL') || die();
 
 require_once(__DIR__ . '/../../src/autoload.php');
-require_once(__DIR__ . '/../../version.php');
+
 use \tool_log\log\writer as log_writer;
 use \tool_log\log\manager as log_manager;
 use \tool_log\helper\store as helper_store;
@@ -89,6 +89,7 @@ class store extends php_obj implements log_writer {
     public function process_events(array $events) {
         global $DB;
         global $CFG;
+        require_once(__DIR__ . '/../../version.php');
         $log_error = function ($message = '') {
             debugging($message, DEBUG_NORMAL);
         };
--- a/src/autoload.php
+++ b/src/autoload.php
@@ -2,6 +2,7 @@
 
 namespace src;
 
+// In Moodle this is set, however in testing this is not. So need to set it if not set or leave it be.
 if (!defined('MOODLE_INTERNAL')) {
     define('MOODLE_INTERNAL', 1);
 }
--- a/src/autoload.php
+++ b/src/autoload.php
@@ -2,7 +2,9 @@
 
 namespace src;
 
-define('MOODLE_INTERNAL', 1);
+if (!defined('MOODLE_INTERNAL')) {
+    define('MOODLE_INTERNAL', 1);
+}
 
 function autoload_src() {
     $directory = new \RecursiveDirectoryIterator(__DIR__);
--- a/src/transformer/events/mod_quiz/attempt_reviewed.php
+++ b/src/transformer/events/mod_quiz/attempt_reviewed.php
@@ -9,7 +9,9 @@ function attempt_reviewed(array $config, \stdClass $event) {
     $learner = $repo->read_record_by_id('user', $event->relateduserid);
     $instructor = $repo->read_record_by_id('user', $event->userid);
     $course = $repo->read_record_by_id('course', $event->courseid);
-    $attempt = $repo->read_record_by_id('attempt', $event->objectid);
+    $attempt = $repo->read_record_by_id('quiz_attempts', $event->objectid);
+    // Quiz attempts don't have names, so this will resolve an issue with the batch send to the LRS later.
+    $attempt->name = 'attempt reviewed';
     $quiz = $repo->read_record_by_id('quiz', $attempt->quiz);
     $lang = utils\get_course_lang($course);
 
--- a/src/transformer/utils/get_full_name.php
+++ b/src/transformer/utils/get_full_name.php
@@ -12,7 +12,7 @@ function get_full_name(\stdClass $user) {
     if ($has_first_name) {
         return $user->firstname;
     }
-    if ($has_lastname_name) {
+    if ($has_last_name) {
         return $user->lastname;
     }
     return '';
--- a/src/transformer/utils/get_user.php
+++ b/src/transformer/utils/get_user.php
@@ -2,9 +2,9 @@
 
 namespace src\transformer\utils;
 
-function get_user(array $config,  \stdClass $user) {
+function get_user(array $config, \stdClass $user) {
     $full_name = get_full_name($user);
-    if ($config['sendmbox'] == true) {
+    if (array_key_exists('sendmbox', $config) && $config['sendmbox'] == true) {
         return [
             'name' => $full_name,
             'mbox' => $user->email,
--- a/version.php
+++ b/version.php
@@ -26,7 +26,7 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin = is_object($plugin) ? $plugin : new StdClass();
+$plugin = isset($plugin) && is_object($plugin) ? $plugin : new StdClass();
 $plugin->component = 'logstore_xapi';
 $plugin->version = 2018061100;
 $plugin->release = '0.0.0-development';
--- a/version.php
+++ b/version.php
@@ -26,7 +26,7 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin = $plugin ? $plugin : new StdClass();
+$plugin = is_object($plugin) ? $plugin : new StdClass();
 $plugin->component = 'logstore_xapi';
 $plugin->version = 2018061100;
 $plugin->release = '0.0.0-development';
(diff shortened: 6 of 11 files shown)

### CODE REVIEW

File: version.php [hunk 1]
```diff
@@ -26,7 +26,7 @@
 
 defined('MOODLE_INTERNAL') || die();
 
-$plugin = $plugin ? $plugin : new StdClass();
+$plugin = isset($plugin) && is_object($plugin) ? $plugin : new StdClass();
```
ryasmi, line 29: Good idea.

src/transformer/utils/get_user.php
```diff
@@ -2,9 +2,9 @@
 
 namespace src\transformer\utils;
 
-function get_user(array $config,  \stdClass $user) {
+function get_user(array $config, \stdClass $user) {
     $full_name = get_full_name($user);
-    if ($config['sendmbox'] == true) {
+    if (array_key_exists('sendmbox', $config) && $config['sendmbox'] == true) {
```
ryasmi, line 7: Nice 👍

src/autoload.php
```diff
@@ -2,7 +2,10 @@
 
 namespace src;
 
-define('MOODLE_INTERNAL', 1);
+// In Moodle this is set, however in testing this is not. So need to set it if not set or leave it be.
```
ryasmi, line 5: Interesting, what do you mean by "in testing"?

classes/log/store.php
```diff
@@ -89,6 +89,7 @@ protected function insert_event_entries(array $events) {
     public function process_events(array $events) {
         global $DB;
         global $CFG;
+        require_once(__DIR__ . '/../../version.php');
```
ryasmi, line 92: Any idea why that works that way but not the other? We have the plugin on a local Moodle 3.5 instance using the other way.

src/autoload.php
caperneoignis: I thought that is why you guys had that there? Otherwise, I don't know why the code needs a redefine for 'moodle_internal'?

classes/log/store.php
caperneoignis: Great question, I'm running 7.0.6 on my local VM and it was throwing errors glore over that. 

Just like the tenery functions for names. ````$site_name = $site->fullname ? $site->fullname : 'A Moodle site';```` if you just leave it as ````$site_name = $site->fullname ?: 'A Moodle site';```` it returns null. Not sure why that is happening, from my understanding that later code line should return $fullname not null.

caperneoignis: Actually, my VM is running windows. What are your dev machines running? There is a chance this is an OS specific issue. In which case, how do you want to handle it? Because this looks like this should work either way, the way I have it now. Windows may have a more strict bug checking because I have xdebug on the windows server.

src/autoload.php
ryasmi: Ah sorry, yeah you're totally right.

classes/log/store.php
ryasmi: Not sure about the ternary's, don't mind changing those ternary's to stop using `?:` that's fine. Just not sure why the `require_once` is causing a problem for that particular file, but not the other one above it `require_once(__DIR__ . '/../../src/autoload.php');`. We're using Mac and Ubuntu mostly here by the way, not sure if that makes a difference.

src/transformer/events/mod_quiz/attempt_reviewed.php
```diff
@@ -9,7 +9,9 @@ function attempt_reviewed(array $config, \stdClass $event) {
     $learner = $repo->read_record_by_id('user', $event->relateduserid);
     $instructor = $repo->read_record_by_id('user', $event->userid);
     $course = $repo->read_record_by_id('course', $event->courseid);
-    $attempt = $repo->read_record_by_id('attempt', $event->objectid);
+    $attempt = $repo->read_record_by_id('quiz_attempts', $event->objectid);
+    // Quiz attempts don't have names, so this will resolve an issue with the batch send to the LRS later.
+    $attempt->name = 'attempt reviewed';
```
ryasmi, line 14: Is the missing attempt name causing an issue on Line 48 in the `context.contextActivities.grouping`? If so I think a better solution would be to make a new util to replace line 48 with something like `utils\get_activity\attempt($config, $attempt, $lang)`. This could then be reused elsewhere to fix the issue in a single function in all uses. Hope that makes sense.

src/transformer/utils/get_full_name.php
```diff
@@ -12,7 +12,7 @@ function get_full_name(\stdClass $user) {
     if ($has_first_name) {
         return $user->firstname;
     }
-    if ($has_lastname_name) {
+    if ($has_last_name) {
```
ryasmi, line 15: Good find, amazed I managed to do that, must have been a dodgy name refactor.

ryasmi (changes requested):

ryasmi:
Really appreciate you taking the time to dig through this @caperneoignis! 👍

src/transformer/utils/get_full_name.php
caperneoignis: Yea, it looked a bit weird when I saw it, thought maybe it was a variable done later, but didn't see it, so I went ahead and changed it to what I thought it should be.

caperneoignis:
No problem @ryansmith94 I need this for a project, so I got time to really sort through this and get it a bit further along for yea.

src/transformer/events/mod_quiz/attempt_reviewed.php
caperneoignis: Yea! That was definitely a fix to get me along to the next issue. Till I found a better way to do it. I will make another util and add it where the $attempt->name is called, I believe it is really just that one return. In which case I'll add the function call there and have a function with a check to see if name is set and if not run it through a switch statement to find the best name depending on the activity table. Shouldn't be too hard and should be much cleaner then what I'm doing now. I'll hit it Monday, if the urge doesn't strike over the weekend.

ryasmi:
That's awesome, really appreciated! 👍

src/transformer/events/mod_quiz/attempt_reviewed.php
ryasmi: No worries, thought as much. Okay sounds good 👍

src/transformer/utils/get_full_name.php
ryasmi: Haha thanks

caperneoignis:
yea, that's fine, I'll start doing the util

PR #176, opened 2018-07-02
added function to find module name and removed temp fix

caperneoignis:
**Description**
- removed temporary fix with something more eligant that can be upgraded later.

**Related Issues**
- #174 

**PR Type**
- Fix and enhancement(will add question submitted as soon as I figure out how to scan for it)

### DIFF
--- a/src/loader/lrs.php
+++ b/src/loader/lrs.php
@@ -25,7 +25,7 @@ function send_batch_to_lrs(array $config, array $statements) {
     $response_text = curl_exec($request);
     $response_code = curl_getinfo($request, CURLINFO_RESPONSE_CODE);
     curl_close($request);
-    
+
     if ($response_code !== 200) {
         throw new \Exception($response_text);
     }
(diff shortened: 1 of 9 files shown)


caperneoignis:
@ryansmith94 This has the get_module_name util function that looks up the name and sets it if it does not exist.

ryasmi:
Thanks @caperneoignis, been a bit stacked today, will try to check this out tomorrow, I think we almost have all of the events working now, so feel free to just plough ahead with the question stuff and we'll get the fixes for the other events pushed up for you tomorrow 👍

ryasmi:
Hey Lee, I'm going to be out of the office from this evening until Monday 16th July so apologies for any slow replies to anything you might do here, although you're also on holiday so probably doesn't matter. If possible do you think you could make a new branch for each "unit of work" (such as finding the module name). That way we can keep merging small chunks. Also it's likely that we may start work on some of the question submission statements while you're away, so it'll hopefully prevent duplicated work.

caperneoignis:
Not a problem, I'll do a different merge request for the get_attempt_responses.

ryasmi:
That's great thanks @caperneoignis 👍

PR #180, opened 2018-07-06
feat: Defaults to using background mode to close #165.

### LINKED ISSUE #165: Cron should be the default way to send statements
Switch the default plugin setting to use the Moodle cron to send statements.

The alternative is blocking execution if Moodle isn't able to handle the load and/or the LRS not able to receive (and respond). The cron would send them as a batch (less load on both Moodle and LRS)

ryasmi:
Hey @davidpesce, just check you're happy with this and my last comment in #165, if so please **squash and merge** this PR using the merge dropdown button below, Github should automatically enter the correct title and description, it should also close issue #165 after the merge is complete. Thank you 👍

### DIFF
--- a/docs/configure-the-plugin.md
+++ b/docs/configure-the-plugin.md
@@ -1,5 +1,7 @@
 # Configure the plugin
 
+**Note that by default the plugin now runs in background mode via a cron task to avoid blocking page responses. This can be changed on the configuration page.**
+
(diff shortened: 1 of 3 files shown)


davidpesce (approved):
Yeah, this is good to go. It will respect the users existing setting if the plugin is currently installed.

PR #182, opened 2018-07-06
feat: Adds setting for sending short course id to close #65.

### LINKED ISSUE #65: Include short id in statement (as an optional setting)
The JISC fork of this plugin includes an extension called http://xapi.jisc.ac.uk/extensions/vle_mod_id which stores the academic short code of the academic course that the Moodle course relates to. This isn't in the best place, and the working group at the LAK hackathon didn't like the identifier because it's not clear what it is. We proposed that this be an object definition extension within the course object using the id https://w3id.org/learning-analytics/learning-management-system/short-id

This would use the Moodle short code, but would ned to be optional as a config setting since not all Moodle instances will use the Moodle short code in this way.

ryasmi:
Hey @davidpesce, just check you're happy with this, if so please **squash and merge** this PR using the merge dropdown button below, Github should automatically enter the correct title and description, it should also close issue #65 after the merge is complete. Thank you 👍

### DIFF
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -96,6 +96,7 @@ class store extends php_obj implements log_writer {
                 'source_version' => $CFG->release,
                 'source_lang' => 'en',
                 'send_mbox' => $this->get_config('mbox', false),
+                'send_short_course_id' => $this->get_config('shortcourseid', false),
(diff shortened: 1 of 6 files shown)


davidpesce (dismissed):
This is also good to go. I really like the addition of is_enabled_config!

davidpesce (approved):

PR #183, opened 2018-07-09
Fix get question submitted

caperneoignis:
**Description**
- initial attempt at adding question responses to statement for quiz submitted, need to fix further

**Related Issues**
- #174 

**PR Type**
- Fix


src/transformer/utils/get_attempt_result.php
```diff
@@ -17,15 +17,17 @@
...
-    $minscore = floatval($gradeitem->grademin ?: 0);
-    $maxscore = floatval($gradeitem->grademax ?: 0);
-    $passscore = floatval($gradeitem->gradepass ?: 0);
+    $minscore = (float)($grade_item->grademin ? $grade_item->grademin : 0);
+    $maxscore = (float)($grade_item->grademax ? $grade_item->grademax : 0);
+    $passscore = (float)($grade_item->gradepass ? $grade_item->gradepass : 0);
 
-    $rawscore = cap_raw_score($gradesum, $minscore, $maxscore);
-    $scaledscore = get_scaled_score($rawscore, $minscore, $maxscore);
+    $rawscore = floatval((cap_raw_score($gradesum, $minscore, $maxscore)));
```
ryasmi: I think we should probably just use `floatval` in the `cap_raw_score` and `get_scaled_score` functions to make this a bit cleaner to read if you agree @caperneoignis

src/transformer/utils/get_module_name.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace src\transformer\utils;
+
+function get_module_name(array $config, $moduletype, $module) {
+    $name = '';
+    if (property_exists($module, 'name')) {
+        $name = $module->name;
+    } else if (property_exists($moduletype, 'name')) {
```
ryasmi: I'm not sure if that will work because I think the module type is usually a string at this point.

caperneoignis: Yes it is a string, that is what we need is a string. The issue is the $module->name does not always exist especially for a quiz, because quizzes do not have a name. So that is why I revert to using $moduletype->name when module does not have a name. It was the fix that allowed mod_quiz events to be properly sent. It was being rejected before with element can not be 'null'

src/transformer/utils/get_attempt_result.php
caperneoignis: Yea, I will see if that works. I think there is a reason I was not doing that, let me try it and see if the issue is resolved. I think it has something to do with the function that is being called. I think that is setting it differently.... But I don't remember which means I should do it again, and verify my findings.

src/transformer/utils/get_module_name.php
ryasmi: Yeah sorry, I'm saying that the `$moduletype` variable is a string not an object, so the name property will never exist. I think you need to change `$moduletype->name` to just `$moduletype` right?

caperneoignis: You are correct, I forgot to change that out after removing the temp fix I used earlier. Sorry about that.

ryasmi: No worries, thanks for checking that Lee 👍

caperneoignis:
@ryansmith94 I think this is much closer to what we need. I also tried to form it so it can be called any where. I'm thinking it can be called with a array_merge with the quiz_submitted statement? But this is probably still missing some items out of the statement. But I'll try and test to see what it looks like.

ryasmi:
Yeah agree that this is close to what we need, it should need to be used with `array_merge` in `attempt_submitted` like you say 👍 Hoping to get around to making an example today.

ryasmi:
Urghhh sorry Lee will have to make the example tomorrow, running out of time today.

PR #194, opened 2018-07-26
feat: Adds statement for essay quiz questions for #174.

ryasmi:
@caperneoignis here's an example of what I was thinking. Part of #174.


caperneoignis:
Aww.... That makes perfect sense. Why didn't I think of that!? Okay, I'll merge when I get a chance. In the process of fixing some other items.

ryasmi:
Glad that makes sense. Haha I'd already thought about this a lot while back. Okay sounds good, cheers Lee.

ryasmi:
@caperneoignis would be good to chat about this on Skype or something to make sure we don't duplicate effort since I've been given some time to work on this next week. Will ping you my details on [Gitter](https://gitter.im).

PR #195, opened 2018-07-26
Add mailto: to the mbox

AndyHubert:
With this, statements get rejected.


ryasmi:
Wow I can't believe we didn't spot that, thanks @AndyHubert 👍

PR #203, opened 2018-07-30
fix: Limits the events processed in the cron to avoid overflows and timeouts.

AndyHubert:
If the backlog in the logstore_xapi_log gets too long, then the call to get_records can either time out or runs out of memory.

### DIFF
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -112,7 +112,7 @@ class store extends php_obj implements log_writer {
                 'lrs_endpoint' => $this->get_config('endpoint', ''),
                 'lrs_username' => $this->get_config('username', ''),
                 'lrs_password' => $this->get_config('password', ''),
-                'lrs_max_batch_size' => get_max_batch_size(),
+                'lrs_max_batch_size' => $this->get_max_batch_size(),
             ],
         ];
         $loadedevents = \src\handler($handlerconfig, $events);
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -77,6 +77,10 @@ class store extends php_obj implements log_writer {
         }
     }
 
+    public function get_max_batch_size() {
+        return $this->get_config('maxbatchsize', 100);
+    }
+
     public function process_events(array $events) {
         global $DB;
         global $CFG;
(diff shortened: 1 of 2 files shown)


classes/task/emit_task.php
```diff
@@ -39,7 +39,7 @@ public function execute() {
         global $DB;
         $manager = get_log_manager();
         $store = new store($manager);
-        $extractedevents = $DB->get_records('logstore_xapi_log');
+        $extractedevents = $DB->get_records('logstore_xapi_log', null, '', '*', 0, $store->get_max_batch_size());
```
ryasmi: Hey @AndyHubert, could you please extract these other arguments (`null`, `''`, `'*'`, and `0`) into variables so it's a little more obvious what they mean? For example the `'*'` might be defined as `$selectedfields = '*'` or something along those lines.

ryasmi (changes requested):
Hey @AndyHubert, thanks for this, really appreciate it. Just one change to request.

classes/log/store.php
```diff
@@ -108,7 +112,7 @@ public function process_events(array $events) {
                 'lrs_endpoint' => $this->get_config('endpoint', ''),
                 'lrs_username' => $this->get_config('username', ''),
                 'lrs_password' => $this->get_config('password', ''),
-                'lrs_max_batch_size' => $this->get_config('maxbatchsize', 100),
+                'lrs_max_batch_size' => $this->get_max_batch_size(),
```
caperneoignis, line 115: Real quick, do these event's get processed from oldest to newest or newest to oldest? Because if it's newest to oldest, the batch size could be small enough, that we start to get a backlog in bigger systems, under heavier usage. So we end up with events, never getting sent, because the amount of new events is greater then the batch size per cron run. Especially with systems seeing 1k users worth of traffic. Just a curious question.

ryasmi: Looking at the code, I believe it would be oldest to newest. I'm going to merge this and we can resolve that issue separately when needed.

ryasmi (approved):

PR #208, opened 2018-07-31
fix: Prevents one transformer error disabling the plugin.

AndyHubert:
If a single statement fails (due to a course or discussion that it refers to having been deleted, for example), then no statements are sent. This same statement comes up first next time and thus the plugin is effectively disabled.

This change allows for a single statement to fail without stopping all others.


ryasmi (approved):
Looks really good thanks @AndyHubert. I might have used `array_reduce` personally to avoid returning null, but I like this just the same. We might need to be wary of how this effects the cron since it'll keep trying to transform the same event in subsequent cron executions, but this is definitely an improvement on the current situation.

PR #209, opened 2018-07-31
fix: Ensures statement duration isn't null.

AndyHubert:
These changes prevents the plugin from crashing when timefinish=0. The `duration` key is not sent in such a case because it cannot be null if sent.


ryasmi (approved):
Nice catch, thanks @AndyHubert. I think we could probably avoid mutating the array to improve readability in case this code grows, but we can address that when the code grows.

PR #228, opened 2018-08-17
feat: Adds support for multichoiceset quiz questions.

caperneoignis:
added an additional case statement to handle multichoiceset qtype. Will need to add a better default handler for future qtypes.

**Description**
- adds an additional handling of one more question type.

**Related Issues**
- didn't make a related issue.

**PR Type**
- Fix

### DIFF
--- a/src/transformer/events/mod_quiz/question_answered/handler.php
+++ b/src/transformer/events/mod_quiz/question_answered/handler.php
@@ -38,6 +38,7 @@ function handler(array $config, \stdClass $event, \stdClass $questionattempt) {
         case 'match':
             return match($config, $event, $questionattempt, $question);
         case 'multichoice':
+        case 'multichoiceset':
             return multichoice($config, $event, $questionattempt, $question);
         case 'numerical':
             return numerical($config, $event, $questionattempt, $question);
--- a/tests/mod_quiz/attempt_submitted/multichoiceset/data.json
+++ b/tests/mod_quiz/attempt_submitted/multichoiceset/data.json
@@ -1,0 +1,71 @@
+{
+    "user": [
+        {
+            "id": 1,
+            "firstname": "test_fullname",
+            "email": "test@test.com"
+        }
+    ],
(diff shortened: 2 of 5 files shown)


caperneoignis:
@ryansmith94 I added an updated qtype handling. I will work on the default handling of future question types later. This is just a quick addition for a qtype I am using and is missing. It should be handled as the multichoice. Since if it is correct or not is handled by Moodle and will display properly if different.

ryansmith94:
I think we should probably create a test for this in `tests/mod_quiz/attempt_submitted` in case someone changes the way this works in future and forgets to add tests. For now it can just be a duplicate of the `tests/mod_quiz/attempt_submitted/multichoice` folder but replace uses of "multichoice" with "multichoiceset".

ryasmi (changes requested):
I think we should probably create a test for this in `tests/mod_quiz/attempt_submitted` in case someone changes the way this works in future and forgets to add tests. For now it can just be a duplicate of the `tests/mod_quiz/attempt_submitted/multichoice` folder but replace uses of "multichoice" with "multichoiceset".

caperneoignis:
Okay I'll update. Do you want me to include the additional changes I have done to the multichoice statement? Since I am adding the correct response and potential answers to it or put that in other branch?

caperneoignis:
nm I will use seperate branch

ryasmi:
Okay can this be closed in favour of #233?

caperneoignis:
@ryansmith94 This branch is just merging the question type, not the answers or the other items, you said you may not want. So this is so the question type all or nothing can be used in the handler. While the other branch is adding the other items that may or maynot be merged.

ryasmi:
Got it, thanks for clarifying 👍

ryasmi (approved):

ryasmi:
Thanks for adding the tests @caperneoignis, just waiting to make sure the build passes before I merge this because I had to merge master into this PR 👍

PR #229, opened 2018-08-18
fix: Validates emails before using them to identify users.

AndyHubert:
This check prevents malformed mbox's from being sent.

### DIFF
--- a/src/transformer/utils/get_user.php
+++ b/src/transformer/utils/get_user.php
@@ -19,8 +19,9 @@ defined('MOODLE_INTERNAL') || die();
 
 function get_user(array $config, \stdClass $user) {
     $fullname = get_full_name($user);
+    $hasValidEmail = filter_var($user->email, FILTER_VALIDATE_EMAIL);
 
-    if (array_key_exists('send_mbox', $config) && $config['send_mbox'] == true) {
+    if (array_key_exists('send_mbox', $config) && $config['send_mbox'] == true && $hasValidEmail) {
         return [
             'name' => $fullname,
             'mbox' => 'mailto:' . $user->email,
--- a/src/transformer/utils/get_user.php
+++ b/src/transformer/utils/get_user.php
@@ -19,9 +19,9 @@ defined('MOODLE_INTERNAL') || die();
 
 function get_user(array $config, \stdClass $user) {
     $fullname = get_full_name($user);
-    $hasValidEmail = filter_var($user->email, FILTER_VALIDATE_EMAIL);
+    $hasvalidemail = filter_var($user->email, FILTER_VALIDATE_EMAIL);
 
-    if (array_key_exists('send_mbox', $config) && $config['send_mbox'] == true && $hasValidEmail) {
+    if (array_key_exists('send_mbox', $config) && $config['send_mbox'] == true && $hasvalidemail) {
         return [
             'name' => $fullname,
             'mbox' => 'mailto:' . $user->email,


src/transformer/utils/get_user.php
```diff
@@ -19,8 +19,9 @@
 
 function get_user(array $config, \stdClass $user) {
     $fullname = get_full_name($user);
+    $hasValidEmail = filter_var($user->email, FILTER_VALIDATE_EMAIL);
```
ryasmi: @AndyHubert can you please rename `hasValidEmail` to `hasvalidemail` in order to pass Moodle's linting rules. Very unfortunate that they don't allow camel-cased variables.

ryasmi (changes requested):
Thanks for this @AndyHubert, very cool, didn't know `filter_var` was able to do this. Just got one small requested change because of Moodle linting rules.

ryasmi (approved):

ryasmi:
Thanks @AndyHubert, sorry for the slow response, been away for a few days.

AndyHubert:
@ryansmith94 This needs to be refined a little, or the mbox validation on the learning locker side should be refined.

I am getting rejected statements again due to the email h*******@hotmail.com being understood as valid on the Moodle plugin side, but invalid by learning locker (and thus the statement was rejected). It was not immediately clear to me where in the LL codebase this email validation is happening.

Options:
1) Correct the mbox validation on LL, since this *is* technically a valid email.
2) Change the mbox validation check behind this PR to match that on LL's side.

Thoughts?

AndyHubert:
The *'s are the special characters. That is, the email is literally `h*******@hotmail.com` in our system. Obviously this is not a real in-use email, but it is legitimately formatted from what I understand and so could theoretically be a real email. Again, the issue is that presently the email validation in the moodle plugin and LL are different and should be unified.

ryasmi:
Ahhhhh haha sorry looked like you had masked the email. The issue with using LL's validation is that it's firstly unlikely to match the validation of other LRSs and secondly that the PHP version of Learning Locker's validation is no longer fully maintained since Learning Locker now uses JavaScript server side.

AndyHubert:
So you don't think it would be wise to conform both validations to some sort of standard? If this is not done, issues will most definitely arise that will unjustly demand both of our time. :)

ryasmi:
I think it would be wise for there to be a standard, but Learning Locker's validation is now written in JavaScript not PHP so it's not usable here. The previous PHP implementation of Learning Locker's validation is no longer maintained because it was rewritten in JavaScript.

It might be worth considering that it's potentially wasteful computationally for us to validate the email twice (once here and once in Learning Locker). We might instead go for a simpler solution here where we rely on Moodle to validate that the email is valid before storing it in the database, then this plugin can simply use the email if it exists and Learning Locker should decide that it's valid too, unless there is a bug in Moodle or Learning Locker's email validation in which case a fix is needed for one of the applications and everyone benefits.

AndyHubert:
Can you point me to where is the code for LL's email validation? I understand it is in JavaScript, but I assume it is using a regex of some sort. If so, I could also just opt to use that same regex for the Moodle plugin instead of the built in PHP email validation. That is the simplest fix in my mind. But I am happy to defer and help if you prefer a different route.

ryasmi:
Hi @AndyHubert, yeah sure, there are two relevant files for Learning Locker's mailto validation (linked below). I'm not sure the current validator would accept the asterix in the email here though actually, so we should really update Learning Locker's JS validator to allow that, there is a [stack overflow post that shows which characters are allowed in the local part of the email address](https://stackoverflow.com/questions/2049502/what-characters-are-allowed-in-an-email-address).

- [The source code with the RegExp](https://github.com/LearningLocker/xapi-validation/blob/master/src/regexValues/mailto.ts)
- [The test code for mailto](https://github.com/LearningLocker/xapi-validation/blob/master/src/tests/regexValues/mailto.ts)

Is Moodle validating these emails before they're stored already?

AndyHubert:
I am not sure what you mean. Before they are stored in the xapi log table? That is not relevant, since it is not until the statements are being prepared that it looks up the user. I will make a fresh pull request.

ryasmi:
No sorry, I mean when the user enters their email into Moodle, does Moodle validate that email address before storing it?

AndyHubert:
Good question. I don't know the answer off hand. The client I am doing work
for has users with invalid emails, but they also use a modified version of
moodle.

ryasmi:
Ah okay that's interesting, can understand why this is needed for that case.

PR #233, opened 2018-08-21
Fix/potential answer mcset

caperneoignis:
**Description**
- adds answers, potential answer, and correct response pattern for multichoice/set qtype

**PR Type**
- Fix

PR #235, opened 2018-08-28
feat: Adds statement for multichoiceset quiz answers.

ryasmi:
This has been branched from PR #233.

@caperneoignis note that I've tidied up a few things rather than making comments on your PR to save some time (such as extracting a function into a util and correctly formatting some JSON in the tests). As you can see below Github has still credited you with the commits 👍 Note that I've added some tests to cover the new setting you've added around sending choices. The build is now failing because we need to decide what the `id` should be for the choices, because currently you've set it to the ID of the answer record (which is a number), but it needs to be a URL to be valid in the xAPI. Let me know what you think we should do about this and I'll update this branch to save making another PR, thanks for all your hard work on this! 😄 

Note that this adds a config setting for sending the available choices and `correctResponsePattern` for multichoice questions.


garemoko:
> The build is now failing because we need to decide what the id should be for the choices, because currently you've set it to the ID of the answer record (which is a number), but it needs to be a URL to be valid in the xAPI.

Do you mean the id for the choices interaction component? That doesn't need to be a URI. See the advice in this blog: https://www.watershedlrs.com/blog/products/xapi-statements-for-quizzes-and-questions (the blog also includes links to the spec)

ryasmi:
Ah thanks @garemoko great spot! Just saw the error was for an ID that looked like an activity and thought it would need to be a URL, but yeah just a string is fine this case, so I'll wrap the answer record id in quotes 👍 I think @caperneoignis had actually already made it a string previously, so totally my bad.

caperneoignis:
Yea, the id is set as a string. I already ran into that issue. Even though a better 'id' could be used to identify the question and description. I have been getting these items directly to my LRS using the settings. I just didn't know how to do the test.

ryasmi:
That's cool. Yeah no worries, we actually have done a test that changes the config anywhere else at this point, usually the change to the config doesn't effect the tests too much, but I felt it would here so decided to switch the config to make sure it was working as expected. We should really test the other config settings more.

caperneoignis:
Actually on the testing note. I think it would be good to figure out a way to dockerize the LRS and have it included in the travisci test. So the unit test can actually be ran against a server, much like the redis, mongo, and other servers are tested against Moodle in the unit test. Obviously there is some difficulty like ensuring the test don't fail due to the docker image for the LRS crashing and making sure the connection can be made. But I think it may be a good idea to test the actual calls against the LRS to ensure the data is showing up there the way we want it too. But that is mostly me thinking aloud.

ryasmi:
Hmmm that's an interesting idea, something we have done on other projects internally.

caperneoignis:
I mean we really would only need the three backend services and not the webUI, just a proxy to send data to those services and ensure the default account information is made a startup through the mongo startup command, or startup script in the LRS container to generate the items needed for connection, kinda like how the admin password is made for Moodle.

ryasmi:
Yeah that's cool, the [Learning Locker xapi-service](https://github.com/LearningLocker/xapi-service) can actually be installed and ran on its own.

PR #239, opened 2018-08-31
fix: Fixes LRS usage with proxy to close #238.

### LINKED ISSUE #238: Moodle proxy setting not used
**Description**
- If the Moodle installation is behind a proxy server, sending statements to an LRS will fail.

**Version**
- observed in 3.14.1 and 3.16.0 

1. Have a proxy configured in $CFG->proxyhost (and possibly $CFG->proxyport).
2. Nothing will be sent as the sent_http_statements function doesn't pass any proxy details to cURL so nothing can be sent.

**Additional information**
- This was previously raised as issue #84 and was implemented at least through the 2.x series. I suspect it has been problematic since the major refactoring that debuted in 3.0.0.

ryasmi:
@peterspicer-catalyst thanks for reporting this #238, I'm unable to test this right now. Are you able to test this by following the instructions below?

1. Navigate to the root directory of your Moodle installation in your terminal.
1. Run the commands below in sequence to get this branch of the plugin.
    ```sh
    cd `admin/tool/log/store`
    rm -rf xapi
    git clone git@github.com:xAPI-vle/moodle-logstore_xapi.git xapi
    cd xapi
    php -r "readfile('https://getcomposer.org/installer');" | php; rm -rf vendor; php composer.phar install
    git checkout origin/fix-proxy
    ```
1. Test that statements are sent from Moodle **with** a proxy configuration.
1. Test that statements are sent from Moodle **without** a proxy configuration.

### DIFF
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -84,7 +84,11 @@ class store extends php_obj implements log_writer {
     private function get_proxy_endpoint() {
         global $CFG;
         if (!empty($CFG->proxyhost)) {
-            return $CFG->proxyhost.':'.$CFG->proxyport;
+            if (!empty($CFG->proxyport)) {
+                return $CFG->proxyhost.':'.$CFG->proxyport;
+            } else {
+                return $CFG->proxyhost;
+            }
         }
         return null;
     }
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -81,6 +81,14 @@ class store extends php_obj implements log_writer {
         return $this->get_config('maxbatchsize', 100);
     }
 
+    private function get_proxy_endpoint() {
+        global $CFG;
+        if (!empty($CFG->proxyhost)) {
+            return $CFG->proxyhost.':'.$CFG->proxyport;
+        }
+        return null;
+    }
+
     public function process_events(array $events) {
         global $DB;
         global $CFG;
@@ -114,6 +122,7 @@ class store extends php_obj implements log_writer {
                 'lrs_username' => $this->get_config('username', ''),
                 'lrs_password' => $this->get_config('password', ''),
                 'lrs_max_batch_size' => $this->get_max_batch_size(),
+                'lrs_proxy_endpoint' => $this->get_proxy_endpoint(),
             ],
         ];
         $loadedevents = \src\handler($handlerconfig, $events);
--- a/src/loader/lrs.php
+++ b/src/loader/lrs.php
@@ -31,6 +31,7 @@ function send_http_statements(array $config, array $statements) {
     $endpoint = $config['lrs_endpoint'];
     $username = $config['lrs_username'];
     $password = $config['lrs_password'];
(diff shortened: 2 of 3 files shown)


ijeffro (approved):

ryasmi:
Hi @peterspicer-catalyst, thanks for the advice on the Moodle Curl Class, do you have a link to the source code for that? I'm having trouble finding it with Google. Once I've had a look at it, I'll determine how feasible it is to use the class itself rather than re-implementing aspects of it. For now it's likely that we'll just re-implement the proxy aspect and refactor to use the entire class later.

peterspicer-catalyst:
Hi @ryansmith94 

Moodle's curl class lives in lib/filelib.php - https://github.com/moodle/moodle/blob/master/lib/filelib.php#L2952 for the class itself in Moodle master.

It's mostly a case of instantiating the curl class, and its post method, where the first parameter is the URL to POST to, the second parameter should be an array of the parameters put into the POST body, and the third parameter should be an array of any extra options you want to pass into cURL itself. Should also give you back status codes etc. for determining if successful.

Peter

ryasmi:
Thanks @peterspicer-catalyst I'm investigating this now. Thanks for the details on how to use the class as well, the only issue I can see with us using that class is that during our automated testing we don't actually have access to the class. Having said that, we don't usually run our tests against the LRS loader, we usually run it against a loader that just returns the statement so that we can check it matches what we'd expect and that it's valid. Using the class for the LRS loader would prevent us from testing the LRS loader locally though when needed, we might be able to create a separate loader that uses the curl class though.

ryasmi:
Just tested that statements are sent from Moodle **without** a proxy configuration. Will try to find a way to test that statements are sent from Moodle **with** a proxy configuration on Monday, then merge this. I also have the `fix-proxy-moodlecurl` branch working, it adds a new loader that uses the Moodle curl, will test that further on Monday too.

peterspicer-catalyst:
Hi @ryansmith94 unfortunately Moodle's docs page doesn't really talk about proxyhost or proxyport, and I think the most comprehensive information to be had on it is actually on the page where you put the values in for Moodle's configuration itself! (If in doubt, use the search function for "proxyhost")

https://docs.moodle.org/35/en/HTTP for the relevant manual page though.

Proxies are something that I'm not particularly versed in, I'm afraid - our sysadmins tell me that we have proxies in front of some clients based on their needs and configurations and I put the relevant details into Moodle and then everything works. Beyond that, hard to say - however as long as the relevant options are passed into cURL, my experience has been that it's pretty transparent.

Peter

ryasmi:
Hi @peterspicer-catalyst, ah that's unfortunate. Thanks for the search tip, I usually use that and Google, just couldn't find anything this time. Thanks for the manual page.

It sounds like by testing that this works without the proxy settings, it should be enough to verify that it should work with the proxy settings since we're using Moodle's curl class.

Moving forward I'll close this PR in favour of #241, since these changes are no longer required to fix #238.

PR #272, opened 2018-09-26
feat: Adds id number in object extension for courses and activities to close #230.

### LINKED ISSUE #230: Adding idnumber to course and activity (course module) extensions
Hi all!

In Moodle all courses and activities have `idnumber` fields. This is useful information to have in xAPI statements as it normally ties back to an external system or record. This reference should persist into the xAPI statement as it directly relates to the learning experience.

I'm currently working on this as it is needed for an internal project however would like to push it back to core. I have a couple of questions about what standards to use and if a merge request would even be accepted.

- Is the key `https://w3id.org/learning-analytics/learning-management-system/` being adopted for all extensions?
short-id has the extension name `https://w3id.org/learning-analytics/learning-management-system/short-id`

- Should this option be behind a toggle flag as short-id is? If this is the standard being set out it a new field group will likely need to be created for optional extra data.

BrendanHalley:
**Description**
Adds id number in object extension for courses and activities under the extension `https://w3id.org/learning-analytics/learning-management-system/external-id`.

There is a config option (disabled by default) for this extension to keep consistency with the shortid extension.

**Related Issues**
Related issue #230

**PR Type**
Enhancement

### DIFF
--- a/src/transformer/utils/get_activity/course.php
+++ b/src/transformer/utils/get_activity/course.php
@@ -22,30 +22,25 @@ use src\transformer\utils as utils;
 function course(array $config, \stdClass $course) {
     $coursename = $course->fullname ? $course->fullname : 'A Moodle course';
     $courselang = utils\get_course_lang($course);
-    $sendshortid = utils\is_enabled_config($config, 'send_short_course_id');
 
-    if ($sendshortid) {
-        return [
-            'id' => $config['app_url'].'/course/view.php?id='.$course->id,
-            'definition' => [
-                'type' => 'http://id.tincanapi.com/activitytype/lms/course',
-                'name' => [
-                    $courselang => $coursename,
-                ],
-                'extensions' => [
-                    'https://w3id.org/learning-analytics/learning-management-system/short-id' => $course->shortname
-                ]
-            ],
-        ];
+    $object = [
+                  'id' => $config['app_url'].'/course/view.php?id='.$course->id,
+                  'definition' => [
+                      'type' => 'http://id.tincanapi.com/activitytype/lms/course',
+                      'name' => [
+                          $courselang => $coursename,
+                      ],
+                  ],
+              ];
+
+    if (utils\is_enabled_config($config, 'send_short_course_id')) {
+        $object['definition']['extensions']['https://w3id.org/learning-analytics/learning-management-system/short-id'] = $course->shortname;
     }
 
-    return [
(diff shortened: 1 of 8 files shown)


src/transformer/utils/get_activity/course.php
```diff
@@ -22,30 +22,25 @@
...
+                      ],
+                  ],
+              ];
+
+    if (utils\is_enabled_config($config, 'send_short_course_id')) {
+        $object['definition']['extensions']['https://w3id.org/learning-analytics/learning-management-system/short-id'] = $course->shortname;
+    }
+
+    if (utils\is_enabled_config($config, 'send_course_and_module_idnumber')) {
+        $courseidnumber = property_exists($course, 'idnumber') ? $course->idnumber : null;
```
ryasmi, line 41: @davidpesce wonder if we should have the same check for `shortname` above, but probably ok.

src/transformer/utils/get_activity/course.php
```diff
@@ -22,30 +22,25 @@
...
+                  'id' => $config['app_url'].'/course/view.php?id='.$course->id,
+                  'definition' => [
+                      'type' => 'http://id.tincanapi.com/activitytype/lms/course',
+                      'name' => [
+                          $courselang => $coursename,
+                      ],
+                  ],
+              ];
+
+    if (utils\is_enabled_config($config, 'send_short_course_id')) {
```
ryasmi, line 36: Usually prefer to avoid mutations because it becomes quite hard to read as the size of the code grows, but that is quite painful in PHP (usually work in JS where this is easier to avoid). Happy with this though, just thought I'd explain why it wasn't like that before, just a personal habit.

ryasmi (approved):
Thanks @trewq, bonus points for the correct PR title and commit format 🌟

tests/xapi_test_case.php
```diff
@@ -77,6 +77,7 @@ protected function get_transformer_config() {
             'send_mbox' => false,
             'send_response_choices' => false,
             'send_short_course_id' => false,
+            'send_course_and_module_idnumber' => false,
```
ryasmi, line 80: @davidpesce this reminds me that we should probably have some tests that check these config values still work in the plugin when they're enabled.

BrendanHalley:
Thanks @ryansmith94. I'm hoping to be able to contribute more.

ryasmi:
You're welcome @trewq, that'd be awesome if you can 👍

PR #281, opened 2018-10-03
fix: Changes email validation to match @LearningLocker.

AndyHubert:
**Description**
- Changes email validation to use the same regex as Learning Locker.

**Related Issues**
- Follows on from #229

**PR Type**
- Fix

### DIFF
--- a/src/transformer/utils/get_user.php
+++ b/src/transformer/utils/get_user.php
@@ -19,7 +19,8 @@ defined('MOODLE_INTERNAL') || die();
 
 function get_user(array $config, \stdClass $user) {
     $fullname = get_full_name($user);
-    $hasvalidemail = filter_var($user->email, FILTER_VALIDATE_EMAIL);
+    // The following email validation matches that in Learning Locker
+    $hasvalidemail = mb_ereg_match("[A-Z0-9\\.\\`\\'_%+-]+@[A-Z0-9.-]+\\.[A-Z]{1,63}$", $user->email, "i");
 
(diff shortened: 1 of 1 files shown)


src/transformer/utils/get_user.php
```diff
@@ -19,7 +19,8 @@
 
 function get_user(array $config, \stdClass $user) {
     $fullname = get_full_name($user);
-    $hasvalidemail = filter_var($user->email, FILTER_VALIDATE_EMAIL);
+    // The following email validation matches that in Learning Locker
+    $hasvalidemail = mb_ereg_match("[A-Z0-9\\.\\`\\'_%+-]+@[A-Z0-9.-]+\\.[A-Z]{1,63}$", $user->email, "i");
```
ryasmi, line 23: The regex here will need a `^` at the beginning to ensure that it matches from the beginning of the string.

ryasmi (changes requested):

ryasmi:
Thanks @AndyHubert, looks good besides that one small point in the regex.

src/transformer/utils/get_user.php
AndyHubert: The [mb_ereg_match](http://php.net/manual/en/function.mb-ereg-match.php) function matches from the beginning of the string automatically.

ryasmi: Ah didn't know that, thanks @AndyHubert, seems good then 👍

ryasmi (approved):

PR #315, opened 2018-11-19
Fix/304 fix html send

### LINKED ISSUE #304: HTML being included in quiz results and response
**Description**
- HTML is being included in quiz results, when the answers and questions are put into the Moodle provided html editor.

1. Create quiz with HTML editor selected in Moodle's editor settings, I forget which one at the moment I'll update when I check the Moodle instance, but I think it happens with either. Add formatting to text, like bold-ing or underlining.
2. Take quiz as a normal non-privileged user.
3. view results in LRS.

**Expected behavior**
- mod_quiz events should be displaying in plain text, or as close to it as possible. Because pulling events for analytics research is more difficult with html in the response string and question string. Harder to find quiz questions when they all start with html tags.

**Actual behaviour**
- HTML tagging is being included in the strings when the events are displaying in the LRS since the html is being included in the string.

**Server information**
- MySQL with normal user signin.

**Client information**
not applicable for this case, since it is happening server to server communication.

**Additional information**
- LRS is LearningLocker, Moodle is 3.5, PHP 7.2, Redis elestic cache, and GlusterFS since all instances are clustered behind ELB.

caperneoignis:
**Description**
- This should resolve #304 it was tested locally using LearningLocker and Moodle and results were coming in after resolving NBSP issue while resolving newline and strip tags issue.

**Related Issues**
- #304 

**PR Type**
- Fix


caperneoignis:
@ryansmith94 Here is the pull request for #304 kind of a pain but I got it working once I figured out the &nbsp html addition that was breaking json_encoding.

caperneoignis:
@ryansmith94 hold on merge, having to correct something, I think a html char is being put in the question text. Otherwise I'm going to have to trace better where the chars are getting placed that will cause the json encoder to crap out since it's not really outputting any errors.

caperneoignis:
@ryansmith94 when using new version 
````
$choices = array_map(function ($answer) use ($lang, $replacestrings) {
            return [
                "id" => "$answer->id",
                "description" => [
                    $lang => str_replace($replacestrings, "", strip_tags($answer->answer))
                ]
            ];
        }, $answers);
````

Is returning an object and not an array, causing the statements to be rejected by the LRS. I'll try and see how I can get around that. But kinda a pain debugging since the exception is never seen.

**Comment by caperneoignis** [comment]
json error output
```
{"errorId":"b3476494-41fd-408d-82f3-f2f6936e8fba","warnings":["Expected 'statements.1.object.definition.choices' to be 'Array'. 
Received '{\"1\":{\"id\":\"1\",
\"description\":{\"en\":\"I had html at one point.\"}},
\"2\":{\"id\":\"2\",\"description\":{\"en\":\"I am not human\"}}}'",
"Expected 'statements.3.object.definition.choices' to be 'Array'. Received '{\"3\":{\"id\":\"3\",\"description\":{\"en\":\"I am human\"}},
\"4\":{\"id\":\"4\",\"description\":{\"en\":\"I am a meat popcicle\"}},\"5\":{\"id\":\"5\",\"description\":{\"en\":\"I am not human\"}}}'"]}
````

**Comment by caperneoignis** [comment]
fixed it, had to add array_values() to strip array items.

**Comment by caperneoignis** [comment]
Well, weird, it works on two of my servers, but the third is still acting like the change hasn't taken place or it's getting some html issue, and there is no error report for the atempt submitted piece on any log, other then the $id thing we have already gone over. So I have not a clue. It works, but I will have to check to see why it does not work on the other server.

File: src/transformer/utils/get_multichoice_definition.php [hunk 1]
```diff
@@ -18,36 +18,42 @@
 defined('MOODLE_INTERNAL') || die();
 
 function get_multichoice_definition(array $config, \stdClass $questionattempt, \stdClass $question, $lang) {
+    // For some reason, newlines and &nbsp is being added to strings,
+    // in order to remove new lines we have to ensure nbsp is also removed.
+    $replacestrings = array("\n", "&nbsp");
```
ryasmi: For consistency with other parts of the code can we change this to the square brackets syntax? This has been done in a few other places too.
```suggestion
    $replacestrings = ["\n", "&nbsp"];
```

src/transformer/utils/get_multichoice_definition.php
```diff
@@ -18,36 +18,42 @@
...
+    // Reducing number of times string replace is called, by moving duplicate code up here.
+    $questiontext = str_replace($replacestrings, "", strip_tags($question->questiontext));
```
ryasmi: Can we please extract this logic into a util function called "cleanHtmlText" or something like that? Thinking we may need it elsewhere and it keeps the transformers a little cleaner with less code duplication.

ryasmi (changes requested):

ryasmi:
Hey @caperneoignis, thanks for this! Just a few changes requested.

caperneoignis:
Good points, should have made a util and was actually thinking it as I was writing it up.

src/transformer/utils/get_multichoice_definition.php
caperneoignis: Yes, doing that now.

ryasmi (approved):
Perrrrrrfick 🐈 Thanks Lee 👍

Issue #317, opened 2018-11-20
attempt_submitted\handler.php is using the wrong table to look up attempt


caperneoignis:
**Description**
- $event->objectid is coming in with 417 for quiz_attempts. However, question_attempts does not have 417. The record is not found and hence no data is being pulled from the record. Some how this is still working unless you have a boat load of attempts. Not sure how it's still working, unless there is an additional lookup deeper for each question and that just so happens to still work with smaller attempt numbers. However, on bigger sites this breaks.
 ````
$questionattempts = $repo->read_records('question_attempts', [
        'questionusageid' => $event->objectid
    ]);
````
**Version**
- my latest branch using the latest master.

**Steps to reproduce the bug**
1. get more quiz attempts then question attempts. Should break the attempt submitted.

**Expected behaviour**
- should pull up the questions for that attempt.

**Actual behaviour**
- nothing is returned for question attempts which means nothing is being sent to the LRS.

**Additional information**
- Since I have to fix this any way, I will try and submit the fix with my question submitted fix. Not sure how this broke, it was working before on the server with a bunch of attempts. I will check file history to see what has changed.

caperneoignis:
This looks like it may solve the issue. 
````
SELECT qa.*, qat.questionid FROM moodle.mdl_quiz_attempts as qa 
join mdl_question_usages as qu on qu.id = qa.uniqueid 
join mdl_question_attempts as qat on qat.questionusageid = qu.id where qa.id = {attemptID};
````
This returns all 15 quiz questions. About to test now.

caperneoignis:
That pretty much was the fix, including these items. $attempt = $repo->read_record('quiz_attempts', array('uniqueid' => $questionattempt->questionusageid)); the quiz_attempts are being sent incorrect ids. I didn't notice this until I started to get super high in attempts for one quiz. Not sure how we haven't had a problem yet. I also had to put back $DB->get_records_sql() so I could pull the complex sql command.

PR #319, opened 2018-11-21
Fix/305 fix record lookup

caperneoignis:
**Description**
- This corrects record lookup issue as listed in #317 

**Related Issues**
- #317

**PR Type**
- Fix


caperneoignis:
@ryansmith94 This corrects the lookup issue and does the lookup using the proper record ids and joins to properly relate question attempt with quiz attempt. This has been tested on the heavy load system and is working. I will double check to ensure lookup is functioning. We may want to add an exception or warning if there are no questions related to a quiz attempt. Since there should be at least one question in a quiz attempt since you can't submit a quiz without any questions. Whether they were answered or not is a different issue. But the questions should still exist.

caperneoignis:
@ryansmith94 I mocked the sql lookup for the test cases because I'm not sure how to mock simulate it, without a massive multi lookup and parsing. Maybe someone over there knows a better way.

src/transformer/events/mod_quiz/question_answered/essay.php
```diff
@@ -24,7 +24,7 @@ function essay(array $config, \stdClass $event, \stdClass $questionattempt, \std
     $repo = $config['repo'];
     $user = $repo->read_record_by_id('user', $event->relateduserid);
     $course = $repo->read_record_by_id('course', $event->courseid);
-    $attempt = $repo->read_record_by_id('quiz_attempts', $questionattempt->questionusageid);
+    $attempt = $repo->read_record('quiz_attempts', array('uniqueid' => $questionattempt->questionusageid));
```
ryasmi: Making this suggestion to keep for consistency with other code.
```suggestion
    $attempt = $repo->read_record('quiz_attempts', ['uniqueid' => $questionattempt->questionusageid]);
```

src/transformer/events/mod_quiz/attempt_submitted/handler.php
```diff
@@ -23,9 +23,12 @@
 
 function handler(array $config, \stdClass $event) {
     $repo = $config['repo'];
-    $questionattempts = $repo->read_records('question_attempts', [
-        'questionusageid' => $event->objectid
-    ]);
+    $sql = "SELECT qat.* FROM {quiz_attempts} as qa
```
ryasmi: I would probably replace this with calls to `$repo->read_record` to simplify tests, it's also likely that we'd want to put some reference to the question usage and question attempt in the statement in future changes.

src/transformer/repos/TestRepository.php
```diff
@@ -48,4 +48,15 @@ public function read_records($type, array $query) {
 
         return $matchingrecords;
     }
+
+    /**
+     * Reads records in a pure sql format. Used for quiz attempt lookup.
+     * @param String $sql
+     * @return mixed|PhpArr
+     */
+    public function read_records_sql($sql) {
```
ryasmi: I think this is likely to cause problems in future developments. Take a loot at one of my earlier comments where I suggest using `read_record` instead. I know it's extra DB calls, but it will allow those other records to be used in future. The extra DB calls should be pretty quick and ran in via the cron anyway so the additional time is quite negligible.

ryasmi (changes requested):

ryasmi:
Hey @caperneoignis, thanks for this PR too. Just made some suggestions to ensure we can build upon this easily going forward.

caperneoignis:
Will update. Good point on using the results from three separate lookups to decrease test complexity I’ll hit that today.

src/transformer/events/mod_quiz/attempt_submitted/handler.php
caperneoignis: removed as requested.

src/transformer/repos/TestRepository.php
caperneoignis: removed as requested in favor of multiple lookups.

src/transformer/events/mod_quiz/attempt_submitted/handler.php
```diff
@@ -23,9 +23,9 @@
 
 function handler(array $config, \stdClass $event) {
     $repo = $config['repo'];
-    $questionattempts = $repo->read_records('question_attempts', [
-        'questionusageid' => $event->objectid
-    ]);
+    $quizattempt = $repo->read_record_by_id('quiz_attempts', $event->objectid);
+    // Other two look ups should be returning one record, This one should return all questions attempted.
+    $questionattempts = $repo->read_records('question_attempts', ['questionusageid' => $quizattempt->uniqueid]);
```
ryasmi, line 28: Nice 👍

src/transformer/events/mod_quiz/question_answered/truefalse.php
```diff
@@ -53,7 +53,8 @@ function truefalse(array $config, \stdClass $event, \stdClass $questionattempt,
             'completion' => $questionattempt->responsesummary !== null,
             'success' => $questionattempt->rightanswer === $questionattempt->responsesummary,
             'extensions' => [
-                'http://learninglocker.net/xapi/cmi/true-false/response' => $questionattempt->responsesummary === 'True',
+                'http://learninglocker.net/xapi/cmi/true-false/response' => $questionattempt->responsesummary ===
+                    'True',
```
ryasmi, line 56: I think this accidentally got moved onto a new line. Shouldn't affect functionality, I'll just move it back when I see it.

ryasmi (approved):

src/transformer/events/mod_quiz/question_answered/truefalse.php
ryasmi: Ah makes sense, it's all good 👍

PR #331, opened 2018-11-30
remove undefined variable $config['lrs_proxy_endpoint'] and unused variable $proxyendpoint

### LINKED ISSUE #327: Error log entries
**Description**
- Upon running in the background, we find the following error emitted:

PHP Notice:  Undefined index: lrs_proxy_endpoint in /path/to/site/admin/tool/log/store/xapi/src/loader/moodle_curl_lrs.php on line 33

**Version**
- 3.17.0

1. Run the plugin - every run of the scheduled task where it emits events to the LRS triggers the error

**Expected behaviour**
- No error should be logged.

**Actual behaviour**
- Error is logged with every call out to cURL.

**Server information**
- N/A

**Client information**
- OS: Linux
- Browser: N/A

**Additional information**
- searching for 'lrs_proxy_endpoint' yields two entries, both of which would trigger errors as both are referring to an array item that doesn't exist.

garemoko:
**Description**
Fixes #327 by removing the lines of code that reference the undefined variable. Since the variable the undefined variable is set to is unused, this should not have any negative consequences. 

Note that proxy LRS functionality is almost certainly broken and this PR makes no attempt to fix it. 

**Related Issues**
- #327

**PR Type**
- Fix


garemoko:
I've tested this code tracking a Moodle quiz to Watershed with developer errors set to maximum and it works.

ryasmi (approved):

ryasmi:
Thanks @garemoko, what makes you say that "proxy LRS functionality is almost certainly broken"?

garemoko:
@ryansmith94 it seemed like it might be because this variable was not being set or used, however I was wrong as confirmed in this comment: https://github.com/xAPI-vle/moodle-logstore_xapi/issues/327#issuecomment-443149019

ryasmi:
Ah okay thanks @garemoko 👍

PR #332, opened 2018-11-30
Fix LRS Proxy

garemoko:
**Description**
- This is an attempt to fix for proxy LRSs based on the code from #239 that was not included in #241. The code added moodle_curlr_lrs.php is new. 

I have not tested with a proxy LRS either before or after these changes so it's possible these changes address an issue that does not exist, or that they do not fix the issue. 

This PR is an alternative to #331 in order to address #327

**Related Issues**
- #239 
- #241 
- #327 
- #331

**PR Type**
- Fix


garemoko:
Closing because apparently the code already worked and this is not needed.  See https://github.com/xAPI-vle/moodle-logstore_xapi/issues/327#issuecomment-443149019

Issue #338, opened 2018-12-05
Multiple ID's used for same verb. 


Ian247:
**Description**
Some cases of different ID's for the same verbs resulting in multiple instances of certain verbs depending on the event being generated. 

**Version**
version 3.18.1

**Steps to reproduce the bug**
1. for completed verb find events that have object.definition.type of "http://id.tincanapi.com/activitytype/legacy-learning-standard" or "http://adlnet.gov/expapi/activities/assessment"
This will have an ID of "http://adlnet.gov/expapi/verbs/completed"
2. compare to type of "http://id.tincanapi.com/activitytype/lms/module" which gives an ID of "http://id.tincanapi.com/verb/completed”
3. Repeat above process for Submitted type "http://id.tincanapi.com/activitytype/survey" gives ID of "http://id.tincanapi.com/verb/submitted"
4. compare this to type "http://adlnet.gov/expapi/activities/assessment" which has a verb ID of "http://activitystrea.ms/schema/1.0/submit"

**Expected behaviour**
- Verb ID's should be consistent to ensure that only having to query one ID for each verb instead of having duplicates. This avoids any issues when building reports. 

**Actual behaviour**
- Duplicate verbs making reporting more difficult

garemoko:
"http://adlnet.gov/expapi/verbs/completed" is the correct completed verb

"http://activitystrea.ms/schema/1.0/submit" is the correct submitted verb

Neither of the http://id.tincanapi.com are listed on https://registry.tincanapi.com

@Ian247 is this a complete list of examples of verbs with multiple ids being used?

PR #340, opened 2018-12-05
Fixes Issue 338 'Multiple ID's used for same verb'

garemoko:
**Description**
- Standardises some verbs as identified in #338. 

Note that this change will lead to data inconsistencies for people that already have data from this plugin, but there are already inconsistencies between events and it's always better to fix sooner rather than later. We should be clear for people to be aware of the implications before upgrading. 

**Related Issues**
- #338

**PR Type**
- Fix


ryasmi (approved):
Thanks @garemoko, is this ready to merge? I'll make sure there is a note in the release that says about the backwards compatibility issues.

ryasmi:
Awesome, thanks for updating the branch. Yeah will squash no worries 👍

PR #362, opened 2018-12-19
Three assignment grade fixes

### LINKED ISSUE #357: assignment score can be negative even when minimum is zero
**Description**
```Warning: Division by zero in /var/www/html/lms/html/admin/tool/log/store/xapi/src/transformer/events/mod_assign/assignment_graded.php on line 56```

One of our clients got the above error when importing historical data. Looking at the code, this could happen if the assignment was scored with a negative score, but the minimum score is zero. I'm not clear on how this could happen, and in particular if Moodle allows assignments to be graded with a score below the minimum or above the max. 

https://github.com/xAPI-vle/moodle-logstore_xapi/blob/master/src/transformer/events/mod_assign/assignment_graded.php#L56

**Version**
Unknown. Recent release from the last month. 

Unknown. Maybe edit an assignment graded database information so it has a negative score and a minimum score of zero. I need to investigate more. 

**Expected behaviour**
Needs more thought. Do we figure out a way to send the statement anyway i.e. by not setting the scaled score? Do we throw a specific error? 

**Actual behaviour**
Row throws an error that's unclear to the end user. 

**Server information**
Unknown

**Client information**
Unknown

**Additional information**
Right after this error, the log shows another error: 

```
Failed load for event id #: {
  "message" : "Missing required argument: statements",
  "status" : 400
}
```
I'm not sure if this is a separate issue, or related, but either way I will log a separate issue as the emitter shouldn't go as far as sending the request to the LRS with a missing statements argument.

garemoko:
**Description**
- Assignment grade values will now not error if the grade lies outside the configured min or max. 
- When an assignment grade is below the passing grade, statements will no longer have an invalid completion value (the string 'unknown'). 
- Assignment graded statements now always have result.completion = true.
- Whether or not the grade awarded is higher than the passing grade is represented by the result.success property. 

Please squash when merging. 

**Related Issues**
- Fixes #357

**PR Type**
- Fix


garemoko:
test failing because:
- result.score.max and result.score.scaled properties are now in a different order. 
- success property is now added.

garemoko:
@ryansmith94 My work is done here. Ready for your review (and merge).

ryasmi (approved):
Looks ok to me, thanks @garemoko 👍

PR #367, opened 2018-12-20
feat: Adds `resendfailedbatches` setting to allow failed statements to be retried.

### LINKED ISSUE #356: Undefined variable: eventobj
**Description**
When logging an error relating to an invalid xapi statement (I'm investigating why the statement is invalid separately) the plugin calls an undefined variable 'eventobj' 

```Notice: Undefined variable: eventobj in /var/www/html/lms/html/admin/tool/log/store/xapi/src/loader/utils/load_batch.php on line 32```

See https://github.com/xAPI-vle/moodle-logstore_xapi/blob/3da6e7532d71b9d46cad16e91eaf18cfaba8fd3b/src/loader/utils/load_batch.php#L32

In addition to the notice, this results in the error not including the event id, making it harder to pin down where the error lies. 

**Version**
Not sure what version our client is running, but I can see the issue in the code in the current master branch.

You would somehow need the plugin to send an invalid xapi statement to an LRS that rejects it, then observe the cron log. 

**Expected behaviour**
There should be no error notice about the undefined variable and the error about the statement should include the event id. 

**Actual behaviour**
There is an error notice about the undefined variable and the error about the statement does not include the event id. 

**Server information**
Unknown

**Client information**
Unknown

**Additional information**
None

### LINKED ISSUE #359: re-send statements one by one if batch fails
In the event of a batch of statements failing, we should re-send the batch one by one. This will mean that good statements in the batch are still sent to the LRS, and will also make debugging the bad statements easier.

### PR DESCRIPTION
**Pull request by garemoko**
**Description**
This PR does a few different things in order to improve the handling of failed batches. 

1. (No related issue) 

Previously emit_task.php would record in the trace a list of event ids which the plugin had attempted to send to the LRS (some or all of which might have been in failed batches) with the message "Events [list of event ids] have been successfully sent to the LRS." But those events had not necessarily been successfully sent and having the event ids served no purpose because the events were no longer in the database in order to look up the ids. 

Now emit_task.php will instead record the number of successful and (separately) unsuccessful events so the person looking at the logs can see how many events were successful and how many were not. 

2. (Fixes #356) 

Previously load_batch.php, in the event of a failed xapi request, would try to trace the event id using the ```$eventobj->id``` variable. However this only resulted in an error because:
A. The ```$eventobj``` variable did not exist in that scope. 
B. It couldn't exist because the xapi request would normally relate to multiple events, not a single id. 
C. Having the event id(s) would have been useless anyway because the database row id for a row that's deleted is useless (the event gets a new id in the failed log table). 

Now the error instead includes the size of the batch that was rejected. 

3. (Fixes #359) 

Previously if a batch failed, all events in that batch were marked as failed, even if there was only one event in the batch that had an issue. 

Now the plugin will recursively retry the events in smaller and smaller batches (half size each time) until it succeeds or gets down to a batch size of one. This means that events will not be rejected for just being in the same batch as a bad request. It also makes debugging of bad events easier because you'll end up with a bad request containing just one event's statement(s). 

Note: I have not tested this with a large data set, only with a pair of events, one of which is bad. 

**Related Issues**
- Fixes #356 
- Fixes #359 

**PR Type**
- Fixes and Enhancement

### CODE REVIEW

**Comment by garemoko** [comment]
@ryansmith94 This and the other two PRs I've opened are ready for review.

File: src/loader/utils/load_batch.php [hunk 1]
```diff
@@ -28,10 +28,23 @@ function load_batch(array $config, array $transformedevents, callable $loader) {
...
         return $loadedevents;
     } catch (\Exception $e) {
+        $batchsize = count($transformedevents);
         $logerror = $config['log_error'];
-        $logerror("Failed load for event id #" . $eventobj->id . ": " .  $e->getMessage());
+        $logerror("Failed load batch (" . $batchsize . " events)" .  $e->getMessage());
         $logerror($e->getTraceAsString());
-        $loadedevents = construct_loaded_events($transformedevents, false);
+
+        // Recursively retry sending statements in increasingly smaller batches so that only the actual bad data fails.
```
ryasmi: I think this will cause some issues because this will delay the time the CRON takes to execute so the next CRON will try to process the same events so we'll duplicate statements. Hope that makes sense.

garemoko: I believe that Moodle cron has some built in stuff that will prevent that happening. See "running cron on multiple servers" here: https://docs.moodle.org/36/en/Cron

ryasmi: Ah didn't know that, good spot.
> Tasks can run in parallel and processes use locking to prevent tasks from running at the same time which allows cron to be triggered from multiple web servers that serve the same Moodle instance.

Do you think it would be better to just validate the statements before we send them? Otherwise if the first statement in the batch is invalid we might send X requests where X is the max statements in a batch.

garemoko: Validating before we send them is good, but it'll be a lot of effort to ensure validation is as good and matches the validation done by the LRS. 

X requests would not be sent in that scenario. Let's say it's a batch of 10 statements the first statement is bad. 

Round 1: 1 batch of 10 statements fails
Round 2: 2 batches of 5 statements (1 fails, 1 succeeds)
Round 3: 2 batches of up to 3 statements (1 fails, 1 succeeds)
Round 4: 2 batches of up to 2 statements (1 fails, 1 succeeds)
Round 5: 2 batches of 1 statement (1 fails, 1 succeeds)

Total requests: 9. Ok, that's actually pretty close to 10, but if you have up to 12 events, it's still only 9 requests! 

I'm open to suggestions of a better approach. The current approach of failing the whole batch is not great.

ryasmi: Yeah matching the validation of the LRS is pretty much impossible given the slight differences between the LRSs and the little holes in the conformance tests. I'd be interested to know what other people think about this issue (@davidpesce and @caperneoignis). I don't think the current approach is too bad, agree that it's not ideal though, but I'm not sure what the ideal solution is either.

garemoko: I'll also add that even if we get the validation perfect, there's still the odd occasion where there's internet gremlins or an LRS has performance struggles (not our LRSs of course!) and a request fails through no fault of the statements. In that case re-sending is a good thing to do too (and we'd only see 2 extra requests in that scenario).

garemoko: I'd be happy to look into making this a setting if people would feel more comfortable with the option to toggle this behaviour.

caperneoignis: Hmm... I do like the idea of weeding out the bad statements so the rest of the statements succeeds. Because nothing is more frustrating then see a bunch of statements missing, and find it it was because one statement had an id wrong or html was in the string, or whatever failure. 

I need to look further into it to see what improvements can be made. I'm more on the "we need to improve error reporting and throws" so we can find issues faster. But also really like the idea of a robust batch send. 

And yes @garemoko is correct, the cli locking is quite nice. But if the item in question gets in a infinity loop that could cause some massive lag in cron, which could cause other issues, just a warning.

garemoko: There shouldn't be any infinite loop with this.

garemoko: Just did some maths and for a batch size of 500 that contains 1 bad statement, it's only 18 additional requests. I'm not sure what volumes of statements are realistic for a regularly running cron on an active Moodle site.

caperneoignis: @garemoko that impressive. That's 3.6% additional request but will allow you to get the 499 other events into the LRS, that is acceptable trade-off. My site generates 5 events from the time a user's login succeeds into the site till they get to the first activitiy. After that, the average is based on what the user is doing. We have a 15 question pre-quiz( 1 event per answered question plus one event for completion, viewing per page, and starting quiz) and a 30 question post quiz, with a tincan_launch activity which will generate about 1.2 event statements per page. So if the user does all three activities, we can be seeing some where in the neighborhood of about 300 events, over 1 to 2 hours. Divide that down to a 5  minute stretch and you are only seeing maybe 5 events, per user, per time block average. 

However, increase that to 1000 concurrent users, and we are seeing about 5K statements per send, give or take. 

I'd almost want to see if we can build in a safe guard for network timeouts, and allow those events to get placed back into the queue and the job just die till next run where it will try the connection again. The other worry, is the job hanging due to a connection timeout or something else, I'd want to put a timeout with a try catch block, Because if a job hangs... I'm not remembering if Moodle handles that gracefully or not or if cron just kind of gets jumbled up and causes a traffic jam. 

FYI: we have our cron running in two minute blocks, with certain jobs schedule for longer windows. But Moodle CRON does allow for locking, so one server will run one thing, the other server will run the other jobs, not locked.

Of course we do our send in real time, because all systems are co-located in the same cloud, so bandwidth is a non-problem and fast. So it's different for my situation.

garemoko: One clarification - the plugin batches events which may generate one or multiple statements. This process is re-running events, not statements, so if the event contains multiple statements and 1 is bad the whole event will still fail. I believe this applies to quiz questions. 

We could make changes to have the task exit in the event of 504 or 503 error codes and send the data later. In that case, you wouldn't want the events to move to the failed events table, you'd just want to leave them where they are. I am cautious about that though as often you would get the 504 and 503 error codes during high traffic periods and not all the requests would fail. Really that's a separate issue anyway because you would want to do that regardless of whether you're re-sending failed requests. I'll create an issue.

garemoko: I made changes so the recursive retry functionality only triggers in the event of a 400 Bad Request response from the LRS. This ensures that in the event of slow responses we're not retrying and retrying so that cron gets blocked. 

I do still think we should retry in the event of 404, 429, 503 and 504, but that can be handled as a separate issue, and probably means retrying next time cron runs rather than in the same cron task.

I like the approach of us tackling specific error codes or groups of error codes one by one, rather than having a catch all approach, so happy for this to start with just 400 errors.

ryasmi: Sorry avoided checking in here to avoid getting in trouble with the better half. I'm quite happy to retry on any error code and perhaps only put events in the failed log if we get a 400-499 error code.

ryasmi: Also putting this behind a flag would be good I think.

garemoko: I added a setting but haven't tested it just yet. I'll comment again once I have tested. (It's passing automated tests)

ryasmi: Cool thanks Andrew 👍

garemoko:
@ryansmith94 this has been tested with both the setting on and off with batches of 2 statements one of which is good and one is bad. In both cases it worked as expected. 

I'm expecting it to pass automated testing and then it should be good for you to review again.

src/loader/utils/load_batch.php
```diff
@@ -28,10 +28,24 @@ function load_batch(array $config, array $transformedevents, callable $loader) {
...
         $logerror = $config['log_error'];
-        $logerror("Failed load for event id #" . $eventobj->id . ": " .  $e->getMessage());
+        $logerror("Failed load batch (" . $batchsize . " events)" .  $e->getMessage());
         $logerror($e->getTraceAsString());
-        $loadedevents = construct_loaded_events($transformedevents, false);
+
+        // In the event of a 400 error, recursively retry sending statements in increasingly
+        // smaller batches so that only the actual bad data fails.
+        if ($batchsize === 1 || $e->getCode() !== 400 || $config['lrs_resend_failed_batches'] !== '1') {
+            $loadedevents = construct_loaded_events($transformedevents, false);
```
ryansmith94: Personally I would return early here instead of assigning to `loadedevents` to be returned later because it's quicker to read since you don't need to scroll down looking for any uses/modifications to `loadedevents`. Happy to merge with this just sharing that because you might agree and find it useful in future.

ryasmi, line 39: Personally I would return early here instead of assigning to `loadedevents` to be returned later because it's quicker to read since you don't need to scroll down looking for any uses/modifications to `loadedevents`. Happy to merge with this just sharing that because you might agree and find it useful in future.

src/loader/utils/load_batch.php
```diff
@@ -28,10 +28,24 @@ function load_batch(array $config, array $transformedevents, callable $loader) {
...
+        $batchsize = count($transformedevents);
         $logerror = $config['log_error'];
-        $logerror("Failed load for event id #" . $eventobj->id . ": " .  $e->getMessage());
+        $logerror("Failed load batch (" . $batchsize . " events)" .  $e->getMessage());
         $logerror($e->getTraceAsString());
-        $loadedevents = construct_loaded_events($transformedevents, false);
+
+        // In the event of a 400 error, recursively retry sending statements in increasingly
+        // smaller batches so that only the actual bad data fails.
+        if ($batchsize === 1 || $e->getCode() !== 400 || $config['lrs_resend_failed_batches'] !== '1') {
```
ryasmi, line 38: I might have flipped this if around and just returned `construct_loaded_events($transformedevents, false);` outside without an `else`, but that is very much a personal preference. I think I prefer that way because it means you don't have to reverse the condition (lacked a better phrase, hope that makes sense). You could instead write the following.

```php
$hasInvalidStatement = $e->getCode() === 400;
$isResendEnabled = $config['lrs_resend_failed_batches'] === '1';

if ($batchsize > 1 && $hasInvalidStatement && $isResendEnabled) {
  // ...
}
```

You could even extract resend logic into a function to remove the need for the comment that currently explains the if block.

```php
$hasInvalidStatement = $e->getCode() === 400;
$isResendEnabled = $config['lrs_resend_failed_batches'] === '1';

if ($batchsize > 1 && $hasInvalidStatement && $isResendEnabled) {
  retry_in_smaller_batches($config, $transformedevents, $loader);
}
```

src/loader/utils/load_batch.php
```diff
@@ -28,10 +28,24 @@ function load_batch(array $config, array $transformedevents, callable $loader) {
...
+            $loadedevents = construct_loaded_events($transformedevents, false);
+        } else {
+            $newconfig = $config;
+            $newconfig['lrs_max_batch_size'] = round($batchsize / 2);
```
ryasmi, line 42: Really like the way you solved this recursively with a modified copy of `$config`.

ryasmi (approved):
Thanks @garemoko. Made some comments that you might find interesting/useful but no need to make any changes.

src/loader/utils/load_batch.php
ryasmi: Yeah sounds good 👍

PR #387, opened 2019-01-07
Book chapter - test case for chapter with no parent

garemoko:
**Description**
I've had a bug report from a client around tracking chapter views which looks like an issue with views of chapters where there is no parent. I realised there are no tests for this so I;m adding tests and then, assuming this PR is failing tests, will fix the issue. 

**Related Issues**
None

**PR Type**
More tests


garemoko:
Replaced with 388

PR #388, opened 2019-01-07
fix book chapter view with no parent

garemoko:
**Description**
Replaces #387 with a cleaner commit history. 

Fixes a bug whereby the read_record_by_id function actually returns numbers as strings and the code was expecting a number for the value of ```$repo->read_record_by_id('book_chapters', $event->objectid)->subchapter```

This is an interesting bug because all of the tests return numbers as numbers, not strings. I have updated the test data for book chapters but not any others. 

**Related Issues**
- #387

**PR Type**
- Fix


garemoko:
Update: returning id values as strings broke the tests, so we'll need to make a more significant change to how the tests work before changing those. For now I've just made it so the non-id number values for book chapter data are represented as strings.

garemoko:
Update 2: It seems using numbers as strings anywhere in the test data causes problems for tests so I've reverted the changes to the tests and modified the code to work regardless of whether the subchapter id is a string or a number.

garemoko:
@ryansmith94 any idea why tests aren't working

garemoko:
@ryansmith94 this is ready for review

ryasmi (approved):
Looks ok.

ryasmi:
Thanks Andrew 👍

PR #446, opened 2019-02-21
Quiz strip html

### LINKED ISSUE #439: Quiz sequencing questions name does not strip html tags.
**Description**
Quiz sequencing questions name does not strip html tags.

**Version**
v3.18.3

Answer a quiz containing a sequencing question (other question types may also be affected)

**Expected behaviour**
- html tags should be stripped from quiz question names because the object,definition.name property is supposed to contain plain text. 

**Actual behaviour**
- html tags are not removed.

garemoko:
**Description**
Currently only true/false and multichoice questions strip html code from the question text. This PR applies that approach to all question types. 

**Related Issues**
Fixes #439

**PR Type**
Fix


garemoko:
Oddly the tests are actually requiring the HTML not to be stripped!

garemoko:
@ryansmith94 this one is ready to review

ryasmi (approved):
Nice 👍

PR #447, opened 2019-02-21
Sequencing response format

### LINKED ISSUE #440: Quiz sequencing questions use wrong response pattern format
**Description**
Quiz sequencing questions use wrong response pattern format

**Version**
v3.18.3

Answer a quiz containing a sequencing question

**Expected behaviour**
The response format should follow the example in the xapi spec. The object.definition.correctResponsesPattern and object.definition.choices properties should also be populated. The http://learninglocker.net/xapi/cmi/sequencing/response extension should not be used. 

See https://github.com/adlnet/xAPI-Spec/blob/master/xAPI-Data.md#sequencing

**Actual behaviour**
```
"result": {
    "success": true,
    "completion": true,
    "response": "{9} {16}",
    "extensions": {
      "http://learninglocker.net/xapi/cmi/sequencing/response": [
        "9",
        "16"
      ]
    }
  }
```

garemoko:
**Description**
Adds response and options information to gap select type questions. 

Note this also includes PR #446 because I was worried about conflicts if I used two separate branches both forked from master.

**Related Issues**
Fixes #440

**PR Type**
- Enhancement


garemoko:
@ryansmith94 this is ready to review

src/transformer/utils/get_multichoice_definition.php
```diff
@@ -32,15 +32,25 @@ function get_multichoice_definition(array $config, \stdClass $questionattempt, \
...
         }, $answers);
+
+        $correctresponsepattern;
+        switch ($interactiontype) {
+            case 'sequencing':
+                $selections = explode('} {', rtrim(ltrim($questionattempt->rightanswer, '{'), '}'));
+                $correctresponsepattern = implode ('[,]', $selections);
+                break;
+            default:
+                $correctresponsepattern = utils\get_string_html_removed($questionattempt->rightanswer);
```
ryasmi: Do choices also need to be imploded in the same way?

garemoko: I'm not sure, but I don't think so because I think the multi-choice question is only one option. 

That would be out of scope for this PR anyway. The goal here is to preserve the behavior of the multi-choice question while changing the gap select one.

ryasmi: I think it's ok to do this for multichoice too because this is [also an issue for multichoiceset questions](https://github.com/xAPI-vle/moodle-logstore_xapi/blob/master/tests/mod_quiz/attempt_submitted/multichoiceset/statements.json).

garemoko: Ok, since you insist, I fixed for multichoice questions where there are multiple correct answers too. :-)

Waiting for tests to run....

garemoko: Tests passed

ryasmi (approved):

src/transformer/utils/get_multichoice_definition.php
ryasmi: Haha okay thanks Andrew 👍 looks good

Issue #271, opened 2018-09-26
Scaled score not calculated properly for quizzes that use marks


petarnikolovski:
**Description**

I have a course that uses both "marks" and "grades" e.g.

![Marks vs Grades - Moodle](https://i.imgur.com/SDLkJHS.png)

But when a user completes a quiz, the quiz score is not calculated properly, instead of getting 1 for the scaled results, a user gets instead mark (raw) converted to percents. In this case: 0.16 (for the last attempt). Also, the attempt was success, and it's accepted, but the success status is set to `false`.

```
    "timestamp": "2018-09-24T10:55:11+01:00",
    "result": {
      "score": {
        "raw": 16,
        "min": 0,
        "max": 100,
        "scaled": 0.16
      },
      "completion": true,
      "success": false,
      "duration": "PT96S"
    }
```

The quiz can have maximum mark of 16, which when scaled to grade is 100%.

**Version**
```
$plugin->version = 2018080302;
$plugin->release = 'v3.14.1';
$plugin->requires = 2014111000;
```

**Steps to reproduce the bug**
1. Download this course https://moodle.net/?coverage=all&language=en&orderby=newest&search=python&downloadable=1&enrollable&submitbutton=1 (Python for Beginners 1 - Python Language Basics)
2. Do any quiz

**Expected behaviour**
- `scaled` should be '1' because user had 16/16 mark.
- `success` should be `true` because user has passed the quiz.

**Actual behaviour**
- `scales` is `0.16`.
- `success` is set to `false`.

**Server information**
- Moodle with Mysql on one server
- LRS is on a separate server running Learning Locker with mongo db

**Client information**
- OS: Ubuntu 16.04

ryasmi:
Hi @petarGitNik thanks for reporting this. Looks like the max score was picked up incorrectly as 100. Could you please share the full statement (removing any personally identifiable information)? The full statement has some additional information in which is useful for us in tracking down the issue.

petarnikolovski:
Hi Ryan, here is the full statement:

```json
{
  "stored": "2018-09-24T10:00:04.323Z",
  "active": true,
  "completedForwardingQueue": [],
  "failedForwardingLog": [],
  "client": "5ba236a44512b2044907712a",
  "lrs_id": "5ba241f24512b2044907712b",
  "completedQueues": [
    "STATEMENT_FORWARDING_QUEUE",
    "STATEMENT_QUERYBUILDERCACHE_QUEUE",
[...]

PR #489, opened 2019-03-30
fix(Quiz Attempt Submitted): Replace raw score with raw grade.

BrendanHalley:
**Description**
The raw score now returned in the xAPI statement is the grade for the attempt. This fixes the issue where scaling was wrong because the matching fields are being used for max and min.

In the past successfully completed questions were returned as the raw value. This change removes no data as it's already in the answered statements for each question.

**Related Issues**
- #271 

**PR Type**
- Fix


garemoko:
Are you able to explain the difference between the two fields i.e. the one currently being used for the score vs. the one that you are changing to?

BrendanHalley:
Old: Score based on a total of correctly answered questions. Moodle defaults to 1 per question.

New: Grade from quiz attempt - Moodle handles all logic based around grading and scaling the grade. This is used for marking so seems most appropriate to be used in this context in the statement.

garemoko:
That makes sense to me.

ryasmi (approved):
Thanks @BrendanHalley, looks good and thanks for updating the tests. Since I don't know Moodle very well, I'm going to have to trust you and the community on this one.

ryasmi:
Hmmm this seems to be failing tests now @BrendanHalley. @garemoko do you think this might have been affected by your changes since I just merged those?

BrendanHalley:
@ryansmith94 The change's merged in include a new test that's missing some data that this pull request requires. I'll fix it up now :)

BrendanHalley:
@ryansmith94 All sorted

ryasmi:
Wow that was fast work, thanks @BrendanHalley 👍

PR #527, opened 2019-05-02
fix(user_report_viewed): Stops errors where course ID is 0.

### LINKED ISSUE #525: user_report_viewed event course id is sometimes 0, but this causes error
**Description**
The user_report_viewed event is triggered when somebody views a report of a particular user's forum usage. This report can look at site wide forum usage rather than for a single course, in which case there is no particular related course. As a result the Moodle event can have a course id of 0, if the user was looking at a report of all forum activity for all courses. 

The generates an error when we try to get course information because course 0 does not exist. 

To fix:
1. Update tests data to have the event include a course id of 0.
2. Delete https://github.com/xAPI-vle/moodle-logstore_xapi/blob/master/src/transformer/events/mod_forum/user_report_viewed.php#L26
and
https://github.com/xAPI-vle/moodle-logstore_xapi/blob/master/src/transformer/events/mod_forum/user_report_viewed.php#L49
3. Update tests expected result. 

I'll try and get a PR in tomorrow. 

**Version**
4.2.4

garemoko:
**Description**
- Fixes https://github.com/xAPI-vle/moodle-logstore_xapi/issues/525
- Adds tests for #525 
- Fixes an issue where the activity id for forum user report views used the current logged in user id, not the id of the user who the report being viewed is about. 

**Known Issues Not Fixed By This PR (Out of Scope)**
If the course if not defined, the lang code defaults to "en". This is the same default that is already used as a fallback in case the lang code of a course is not valid, so I don't feel bad about using this default again here. Ideally we should use the site default as the fallback, but that should be fixed everywhere and is a separate issue. 

**Related Issues**
- #525 

**PR Type**
- Fix


garemoko:
@ryansmith94 This is ready for review.

src/transformer/utils/get_activity/user_report.php
```diff
@@ -19,20 +19,26 @@
...
-    $courselang = utils\get_course_lang($course);
+function user_report(array $config, \stdClass $user, \stdClass $course, $relateduserid, $courselang) {
 
-    return [
-        'id' => $config['app_url'].'/mod/forum/user.php?id='.$user->id.'&course='.$course->id,
+    $activity = [
         'definition' => [
             'type' => 'http://id.tincanapi.com/activitytype/user-profile',
             'name' => [
                 $courselang => 'forum posts of '.utils\get_full_name($user),
```
ryasmi, line 28: Since we're using the related user ID below for the activity ID, does that mean we should use the related user here? Same comment for the extension below too.

ryasmi: @garemoko

garemoko: Yes, definitely. Well spotted!

garemoko: In which case i probably shouldn't have modified the user part of this file, but instead should have just changed the user that is passed in by user_report_viewed.php

ryasmi: Sweet. Yeah good idea that probably makes the most sense.

garemoko: @ryansmith94 this is resolved.

ryasmi (approved):
Noice 👍

Issue #571, opened 2019-06-04
Totara program 'enrolled to' statement


isheeler:
**Description**
We are using the plugin on a Totara installation. Totara has a 'program' feature that groups courses together. Learners who enrol on one course are enrolled in all courses within that program. It would be great if the plugin could send statements to report when someone enrolled on a program so that we can keep a count of unique learners on a program. 

**Version**
3.0

**Steps to reproduce the bug**
It's Totara specific. No way to send a statement of enrolment on a program. At the moment we can only report when someone enrols on a course within that program, and this leads to people being counted multiple times.

**Expected behaviour**
Learner enrolled to [programe name]  

**Actual behaviour**
None

BrendanHalley:
This brings up an interesting point about whether Totara events are going to be included in this plugin or this project forked.

I've been meaning to add all the unique Totara events to this plugin for a while.

davidpesce:
I think it may be confusing for users to have totara-specific admin settings. There are already enough events in the list that I'm thinking it needs to be a searchable/filtered.

That being said, rather than forking for a new plugin, why not a separate branch on this repo?

BrendanHalley:
Seperate branch would work fine. I work for a Totara partner and I'm happy to maintain the branch.

I'll get an issue up with the proposed way forward for this plugin and Totara.

ryasmi:
I agree with David that "it may be confusing for users to have totara-specific admin settings". I also agree that the events list might need to be searchable (that one is probably a separate issue).

I think I would prefer that the project wasn't forked, I would also prefer that we didn't have a separate branch. I think I would prefer to have a moodle folder and a totara folder, but I think that might complicate installation and our CI process because I think certain files are expected in certain folders. Therefore, whilst the separate branch is not ideal, I think it's probably the way we'll have to go.

garemoko:
Coming at this from another angle, this could be seen as just an example of a more general problem that the Moodle site might not have events that this plugin tracks. In addition to the Totora case, other cases might include:

- We add tracking for an event associated with an non-core plugin
- An event is removed in a later version of Moodle 
- A Moodle site which uninstalls some core modules that are not required in order to simplify the Moodle.

I wonder if it's possible for us to detect which events are registered on the current Moodle site and only display settings for those events. See https://github.com/moodle/moodle/blob/6153be6850869cdc3a6ae925dcf6e688ac481333/report/eventlist/classes/list_generator.php#L52

[...]

PR #583, opened 2019-06-12
feat(Totara): Adds `\totara_program\event\program_assigned` event.

BrendanHalley:
**Description**
It is necessary for us to have a mechanism that allows an event to be supported by the plugin but is not supported by the environment it's installed on.

Moodle has a core function that lists all events, our mapping function should only return events that the environment has installed and the plugin supports.

`\totara_program\event\program_assigned` has also been added as an example of how Totara events will be implemented.

**Related Issues**
#571

**PR Type**
Enhancement


src/transformer/utils/extensions/base.php
```diff
@@ -17,7 +17,7 @@
 namespace src\transformer\utils\extensions;
 defined('MOODLE_INTERNAL') || die();
 
-function base(array $config, \stdClass $event, $course) {
+function base(array $config, \stdClass $event, $course=null) {
```
garemoko, line 20: I'm curious about the reason for this change.

src/transformer/utils/totara/program.php
```diff
@@ -0,0 +1,34 @@
...
+namespace src\transformer\utils\totara;
+defined('MOODLE_INTERNAL') || die();
+
+function program(array $config, \stdClass $program, $lang) {
+    $programname = $program->fullname ? $program->fullname : 'A Totara program';
+
+    $object = [
+        'id' => $config['app_url'].'/totara/program/view.php?id='.$program->id,
+        'definition' => [
+            'type' => 'http://id.tincanapi.com/activitytype/lms/program',
```
garemoko, line 26: This activity type is not registered at https://registry.tincanapi.com/ and is not hosted on id.tincanapi.com. 

http://id.tincanapi.com/activitytype/learning-program is also not on registry, but I believe I have submitted it for consideration and I've definitely used it.

src/transformer/utils/extensions/base.php
BrendanHalley: I've used `base` in `src/transformer/events/totara_program/program_assigned.php` in a context where a course doesn't exist. The course parameter is only required in this function for the jisc data.

In `src/transformer/utils/extensions/course_area.php` the function `course_area` (called from jisc util) already has a condition for if the course is null. I expect my change was actually how it was intended to be initially.

src/transformer/utils/totara/program.php
BrendanHalley: In this plugin we have used `http://id.tincanapi.com/activitytype/lms/course` and `http://id.tincanapi.com/activitytype/lms/module`. Keeping with Programs being a function of the LMS I thought `http://id.tincanapi.com/activitytype/lms/program` would be appropriate.

Happy to change it to `http://id.tincanapi.com/activitytype/learning-program` though if that's what you recommend, I was just going for consistency.

src/transformer/utils/extensions/base.php
garemoko: Makes sense thanks for explaining

src/transformer/utils/totara/program.php
garemoko: Consistency is important. 

@ryansmith94 @davidpesce any idea why http://id.tincanapi.com/activitytype/lms/course and http://id.tincanapi.com/activitytype/lms/module are used since these do not exist on http://id.tincanapi.com ?

ryasmi: I can't recall why they were used, seems likely that they were based off something else that did/does exist on the registry.

BrendanHalley: Is there any chance we can get this split off into a separate issue so the Totara event can be merged?

ryasmi: Agreed that the course and module activity types should be discussed in a separate issue. We should probably decide what the activity type should be for the learning program though to avoid having to change that too.

ryasmi: Since a "program" is quite a Totara specific activity, perhaps we should make a Totara specific activity type, what do you think @garemoko?

garemoko: I think it's a familiar concept in a few LMSs. I have used http://id.tincanapi.com/activitytype/learning-program before and I believe that's already in the queue for approval on registry.

ryasmi: Okay cool

ryasmi (approved):

Issue #604, opened 2019-08-30
Missing privacy provider, causing failing unit tests


yao9394:
**Description**
- missing privacy provider is making Moodle complain on unit tests

**Version**
- master at 2d1fd066923dedcdb7368b83fda5c8fc2ea4dcb3 on 2018082100

**Steps to reproduce the bug**
1. run command ```vendor/bin/phpunit "provider_testcase" privacy/tests/provider_test.php```

**Expected behaviour**
- should see no failing tests

**Actual behaviour**
- we see:
```1) provider_testcase::test_all_providers_compliant with data set "logstore_xapi" ('logstore_xapi', 'logstore_xapi\privacy\provider')
Failed asserting that false is true.

/siteroot/privacy/tests/provider_test.php:178
/siteroot/lib/phpunit/classes/advanced_testcase.php:80

To re-run:
 vendor/bin/phpunit "provider_testcase" privacy/tests/provider_test.php

2) provider_testcase::test_table_coverage
The following tables with user fields must be covered with metadata providers: 
  - plagiarism_turnitin_users (userid)
  - logstore_xapi_log (userid)
  - logstore_xapi_failed_log (userid)
```

**Client information**
- OS: Linux

davidpesce:
Closing in favor of the new feature issue #617

PR #605, opened 2019-08-30
Add privacy provider to the plugin

yao9394:
**Description**
- This PR is to add privacy provider to the plugin

**Related Issues**
- #604 

**PR Type**
- Fix

### DIFF
--- a/classes/privacy/provider.php
+++ b/classes/privacy/provider.php
@@ -1,0 +1,120 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


davidpesce:
@yao9394 - thanks for submitting this! it looks as if it's the skeleton for the privacy API rather than a full implementation. Do you have any time to fill in the details for the following:
* Indicating that you export data to an external location
* Providing a way to export user data
* Retrieving the list of contexts
* Deleting user data

If not, I'll merge this as is and create a new issue where this can be added later.

davidpesce (approved):
I'm going to create a new issue with the specifics needed to fully implement the privacy api

PR #612, opened 2020-01-24
Fixing ci issues

davidpesce:
Fixing issues with ci

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: php
 sudo: false
 
+services:
+  - mysql
+  - postgresql
+
 cache:
   directories:
     - $HOME/.composer/cache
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,21 +1,24 @@
 language: php
 sudo: false
 
+services:
```
ryasmi, line 4: Do we actually need these services running? I don't think we actually hit the DB.

ryasmi (commented):
I think this looks ok to me, just a comment on the services.

.travis.yml
davidpesce: It's required if we use the matrix values here:

  matrix:
    - DB=pgsql
    - DB=mysqli

And those are required by the moodle-plugin-ci

ryasmi: Okay no worries.

ryasmi (approved):

PR #620, opened 2020-02-04
bumping php version

davidpesce:


davidpesce:
@ryansmith94 - it doesn't look like we can bump this just yet. Any chance you can fix this in the statement factory? https://github.com/LearningLocker/StatementFactory/issues/16

ryasmi:
Ah that's annoying, yeah we can probably fix this in there.

PR #621, opened 2020-02-04
fixes composer errors and bumps php version to something zend support…

davidpesce:
fixes composer errors and bumps php version to something zend supports (php 7.2)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,10 +72,10 @@ deploy:
     skip_cleanup: true
     on:
       tags: true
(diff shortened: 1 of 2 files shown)


davidpesce:
@ryansmith94 - hoping this will fix our lockfile maintenance. I can't bump it to php 7.3 because moodle 3.5 doesn't support it fully.

ryasmi (approved):
Thanks @davidpesce 👍

Issue #454, opened 2019-03-05
misleading wording in mtrace message


garemoko:
The message "X event(s) have been successfully sent to the LRS" is traced when running the cron task. This is misleading because many events are successfully processed but actually do not send any data to the LRS (i.e. stuff that Moodle logs that we haven't yet implemented tracking for). 

Suggested new wording: "X event(s) have been successfully processed"

For unsuccessfully processed events, "X event(s) have failed to send to the LRS" is probably accurate, since I don't think we can fail on events that do not attempt to send data to the LRS.

ryasmi:
Yeah good catch, sorry I didn't reply sooner.

davidpesce:
Fixed in PR #627

PR #627, opened 2020-02-06
fix: mtrace discrepancy per issue #454 

davidpesce:
Fixed mtrace discrepancy per issue #454 and moved them to lang file for alt…ernate languages


ryasmi (commented):
Nice change.

ryasmi (approved):

PR #673, opened 2020-04-01
Lms 1655 1656 config option add cohort to notifications

mlynn-lp:
**Description**
- Add config options to add cohorts and additional email addresses to notifications.

**Related Issues**
- #1655 and 1656

**PR Type**
- Enhancement


davidpesce:
@mlynn-lp - I think there's an issue with this. When I do the install, it gets to the point where I adjust new settings and click save. I think this usually occurs when there is an admin setting that isn't set in the db and the plugin is attempting to set it. 

Any chance you know which one may be failing? Whichever one it is, it'll need to be added to the upgrade.php

davidpesce (changes requested):
Are there missing table values? I'm having an issue when upgrading to this version from the previous version. The plugin installs and prompts me to change the new settings. Clicking save brings me back to the same page with nothing available to change. Essentially stuck in a loop. See below.

Any ideas?

image

settings.php
```diff
@@ -67,6 +83,35 @@
...
+    // Cohorts.
+    $settings->add(new admin_setting_heading('cohorts',
+        get_string('cohorts', 'logstore_xapi'),
+        get_string('cohorts_help', 'logstore_xapi')));
+
+    $cohorts = logstore_xapi_get_cohorts();
+    $arrcohorts = array();
+    foreach ($cohorts as $cohort) {
+        $arrcohorts[$cohort->id] = $cohort->name;
+    }
```
davidpesce, line 105: Figured it out! It's because I didn't have any cohorts on the Moodle instance. Should probably account for this here.

settings.php
```diff
@@ -67,6 +83,35 @@
...
+
+    $settings->add(new admin_setting_configmulticheckbox('logstore_xapi/cohorts',
+        get_string('includecohorts', 'logstore_xapi'), '', '', $arrcohorts));
+
+    // Additional email addresses.
+    $settings->add(new admin_setting_configtext('logstore_xapi/send_additional_email_addresses',
+        get_string('send_additional_email_addresses', 'logstore_xapi'),
+        get_string('send_additional_email_addresses_desc', 'logstore_xapi'), '', PARAM_TEXT));
+
```
davidpesce: Figured it out! It's because I didn't have any cohorts on the Moodle instance. Should probably account for this here.

mlynn-lp: Hi David, I have changed my code to not display this setting if no cohorts are setup on the Moodle instance. This should fix the problem during db upgrades.

mlynn-lp: Sorry David, I think this should be being merged into master-jisc not master. I will discuss this with Gordon in the morning.

davidpesce: No worries. It looks like Gordon merged this into master-jisc yesterday. We chatted over email, but always better to check.

PR #679, opened 2020-04-08
LMS-1693: new scheduled task to replay statements

mlynn-lp:
**Description**
- Add ability to resend failed requests with emit_failed scheduled task.

**Related Issues**
- #LMS-1693

**PR Type**
- Enhancement

### DIFF
--- a/classes/task/failed_task.php
+++ b/classes/task/failed_task.php
@@ -31,37 +31,55 @@ class failed_task extends \core\task\scheduled_task {
         return get_string('taskfailed', 'logstore_xapi');
     }
 
-    private function extract_failed_events($limitnum) {
+    /**
+     * Extract events from logstore_xapi_log or logstore_xapi_failed_log.
+     *
+     * @return array
+     */
+    private function extract_events($limitnum=0, $type='log') {
         global $DB;
+
         $conditions = null;
         $sort = '';
         $fields = '*';
         $limitfrom = 0;
-        $extractedevents = $DB->get_records('logstore_xapi_failed_log', $conditions, $sort, $fields, $limitfrom, $limitnum);
-        return $extractedevents;
+
+        if ($type == 'log') {
+            $events = $DB->get_records('logstore_xapi_log', $conditions, $sort, $fields, $limitfrom, $limitnum);
+        } else {
+            $events = $DB->get_records('logstore_xapi_failed_log', $conditions, $sort, $fields, $limitfrom, $limitnum);
+        }
+        return $events;
     }
 
-    private function extract_events($limitnum) {
-        global $DB;
-        $conditions = null;
-        $sort = '';
-        $fields = '*';
-        $limitfrom = 0;
-        $extractedevents = $DB->get_records('logstore_xapi_log', $conditions, $sort, $fields, $limitfrom, $limitnum);
-        return $extractedevents;
-    }
-
+    /**
+     * Insert events into logstore_xapi_log.
+     *
+     * @return none
+     */
     private function insert_failed_events_into_xapi_log($events) {
         global $DB;
+
         $DB->insert_records("logstore_xapi_log", $events);
     }
 
+    /**
+     * Delete events from logstore_xapi_failed_log.
+     *
+     * @return none
+     */
     private function delete_failed_events($events) {
         global $DB;
+
         $eventids = $this->get_delete_event_ids($events);
         $DB->delete_records_list('logstore_xapi_failed_log', 'id', $eventids);
     }
 
+    /**
+     * Get failed events and add in errortype and response.
+     *
+     * @return array
+     */
     private function get_failed_events($events) {
         $nonloadedevents = array_filter($events, function ($loadedevent) {
             return $loadedevent['loaded'] === false;
@@ -70,7 +88,7 @@ class failed_task extends \core\task\scheduled_task {
             return $nonloadedevent['event'];
         }, $nonloadedevents);
 
-        // enrich failed event with errortype and response
+        // Enrich failed event with errortype and response.
         foreach ($failedevents as $event) {
             if (!empty($_SESSION['xapi_errortype'])) {
                 $event->errortype = $_SESSION['xapi_errortype'];
@@ -82,6 +100,11 @@ class failed_task extends \core\task\scheduled_task {
         return $failedevents;
     }
 
+    /**
+     * Get successful events.
+     *
+     * @return array
+     */
     private function get_successful_events($events) {
         $loadedevents = array_filter($events, function ($loadedevent) {
             return $loadedevent['loaded'] === true;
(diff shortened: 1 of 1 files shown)


classes/task/failed_task.php
```diff
@@ -31,6 +31,98 @@ public function get_name() {
         return get_string('taskfailed', 'logstore_xapi');
     }
 
+    private function extract_failed_events($limitnum) {
+        global $DB;
+        $conditions = null;
```
lzabo: Code styling: Could you add extra empty line after globals, please?

classes/task/failed_task.php
```diff
@@ -31,6 +31,98 @@ public function get_name() {
         return get_string('taskfailed', 'logstore_xapi');
     }
 
+    private function extract_failed_events($limitnum) {
```
lzabo: Could you use protected instead of private?

lzabo: Could you add php doc blocks?

classes/task/failed_task.php
```diff
@@ -31,6 +31,98 @@ public function get_name() {
...
+
+    private function get_failed_events($events) {
+        $nonloadedevents = array_filter($events, function ($loadedevent) {
+            return $loadedevent['loaded'] === false;
+        });
+        $failedevents = array_map(function ($nonloadedevent) {
+            return $nonloadedevent['event'];
+        }, $nonloadedevents);
+
+        // enrich failed event with errortype and response
```
lzabo: code styling issue at Inline comments.
Inline comments must use the "// " (2 slashes + whitespace) style, laid out neatly so that it fits among the code and lines up with it. The first line of the comment must begin with a capital letter (or a digit, or '...') and the comment must end with a proper punctuation character. Permitted final characters are '.', '?' or '!'.

classes/task/failed_task.php
```diff
@@ -31,6 +31,98 @@ public function get_name() {
...
+        }, $loadedevents);
+    }
+
+    private function get_delete_event_ids($loadedevents) {
+        $arr = array();
+        foreach ($loadedevents as $event) {
+            $arr[] = $event->id;
+        }
+        return $arr;
+    }
```
gordonmacqueen-lp: what is the need for the `get_delete_event_ids` function @mlynn-lp, doesn't the above `get_event_ids` do the same?

classes/task/failed_task.php
```diff
@@ -31,6 +31,98 @@ public function get_name() {
         return get_string('taskfailed', 'logstore_xapi');
     }
 
+    private function extract_failed_events($limitnum) {
+        global $DB;
+        $conditions = null;
+        $sort = '';
+        $fields = '*';
+        $limitfrom = 0;
+        $extractedevents = $DB->get_records('logstore_xapi_failed_log', $conditions, $sort, $fields, $limitfrom, $limitnum);
```
lzabo: if I understand well this one similar with:
  $extractedevents = $DB->get_records('logstore_xapi_failed_log', null, null, null, null, $limitnum);

We definitely don't need any extra declaration here because all of them are set in the function declaration.

classes/task/failed_task.php
```diff
@@ -31,6 +31,98 @@ public function get_name() {
...
+        return $extractedevents;
+    }
+
+    private function extract_events($limitnum) {
```
lzabo: if we add second parameter which is the table name or something similar to the function we can use only one function instead of extract_failed_events and extract_events.
I mean for example:
```
    protected function extract_events($limitnum=0, $type='failed') {
        global $DB;

        if ($type == 'failed') {
            $events = $DB->get_records('logstore_xapi_failed_log', null, null, null, null, $limitnum);
        } else {
            $events = $DB->get_records('logstore_xapi_log', null, null, null, null, $limitnum);
        }
        return  $events;
    }

```

classes/task/failed_task.php
lzabo: Please declare default value if possible. In this case, 0 should be good.

lzabo (changes requested):

classes/task/failed_task.php
mlynn-lp: I have fixed this and will commit the change later.

classes/task/failed_task.php
mlynn-lp: array_map throws a warning when the closure raises an exception.

classes/task/failed_task.php
mlynn-lp: This was copied from the pre-existing emit_task.php

classes/task/failed_task.php
mlynn-lp: I have fixed this and will commit today.

classes/task/failed_task.php
mlynn-lp: I think the original programmer wanted to be able to change this easily if necessary, so the parameters are declared explicitly.

classes/task/failed_task.php
mlynn-lp: I have combined the methods into one as recommended but the parameters cannot be null, null, null, null. They must have suitable defaults because the limit is configurable and passed by parameter. For easy configuration I have kept the original explicit declarations.

 ```
   private function extract_events($limitnum=0, $type='log') {
        global $DB;

        $conditions = null;
        $sort = '';
        $fields = '*';
        $limitfrom = 0;

        if ($type == 'log') {
            $events = $DB->get_records('logstore_xapi_log', $conditions, $sort, $fields, $limitfrom, $limitnum);
        } else {
            $events = $DB->get_records('logstore_xapi_failed_log', $conditions, $sort, $fields, $limitfrom, $limitnum);
        }
        return $events;
    }
```

Comment (lzabo) [reply classes/task/failed_task.php]: if you rename $type variable to $table
that you can only write the following:
```
    private function extract_events($table='logstore_xapi_log', $limitnum=0) {
        global $DB;

        $conditions = null;
        $sort = '';
        $fields = '*';
        $limitfrom = 0;
        $events = array();
 
        if ($table == 'logstore_xapi_log' || $table == 'logstore_xapi_failed_log') {
            $events = $DB->get_records($table, $conditions, $sort, $fields, $limitfrom, $limitnum);
        }
        return $events;
    }
```

What do you think?

lzabo: and the call looks like for example:
`$this->extract_events('logstore_xapi_log', 5);`

lzabo: readable, isn't it?

mlynn-lp: The condition if($table == 'logstore_xapi_log' || $table == 'logstore_xapi_failed_log') is too long since we know the tables already. If we were separating out the database layer we would never pass a table name to the method. Also this would involve 2 string compares each time the method is called.

So we don't have to pass the table name and to keep the definition for the method the same for the default table I think I will keep the $type parameter instead.

classes/task/failed_task.php
lzabo: make sense.

lzabo (approved):

gordonmacqueen-lp (approved):

classes/task/failed_task.php
gordonmacqueen-lp: In that case, I'm not sure we need both functions.
Happy to leave for now

Issue #344, opened 2018-12-06
\mod_scorm\event\sco_launched in failed log


ryasmi:
**Version**
- 3.17.0

**Steps to reproduce the bug**
1. Run the plugin with `\mod_scorm\event\sco_launched` events in the xapi log.
2. Observe the failed log.

**Expected behaviour**
- Event should be emitted and removed from the xapi log.

**Actual behaviour**
- Event is not emitted, it is removed from the xapi log, but it is added to the failed log.

**Server information**
- Postgres on Ubuntu 14.04 LTS

mlynn-lp:
The objectid recorded in mdl_logstore_standard_log for sco_launched activities is the scoid.

This is incorrectly used here:
`$scorm = $repo->read_record_by_id('scorm', $event->objectid);`
This code throws an exception if the $event->objectid is not found.

Affected file:
admin/tool/log/store/xapi/transformer/events/mod_scorm/sco_launched.php

Suggested solution:
```
$scormscoes = $DB->get_record('scorm_scoes', array('id' => $event->objectid), 'scorm');
$scorm = $repo->read_record_by_id('scorm', $scormscoes->scorm);
```

ryasmi:
Thanks @mlynn-lp 👍 tagging @gordonmacqueen-lp

davidpesce:
Fix merged in #686

PR #686, opened 2020-04-15
fix sco_launched transformation

mlynn-lp:
**Description**
- The sco_launched SCORM message would fail transformation because the objectid referred to the scoid not the row in the scorm table. This fixes the issue.

**Related Issues**
- #344

**PR Type**
- Fix

### DIFF
--- a/src/transformer/events/mod_scorm/sco_launched.php
+++ b/src/transformer/events/mod_scorm/sco_launched.php
@@ -21,12 +21,10 @@ defined('MOODLE_INTERNAL') || die();
 use src\transformer\utils as utils;
 
 function sco_launched(array $config, \stdClass $event) {
-    global $DB;
-
     $repo = $config['repo'];
     $user = $repo->read_record_by_id('user', $event->userid);
     $course = $repo->read_record_by_id('course', $event->courseid);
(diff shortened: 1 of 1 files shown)


src/transformer/events/mod_scorm/sco_launched.php
```diff
@@ -21,10 +21,13 @@
 use src\transformer\utils as utils;
 
 function sco_launched(array $config, \stdClass $event) {
+    global $DB;
+
     $repo = $config['repo'];
     $user = $repo->read_record_by_id('user', $event->userid);
     $course = $repo->read_record_by_id('course', $event->courseid);
-    $scorm = $repo->read_record_by_id('scorm', $event->objectid);
+    $scormscoes = $DB->get_record('scorm_scoes', array('id' => $event->objectid), 'scorm');
```
ryasmi: This will fail tests, you should fetch records using the `$repo` variable instead. The tests use a different repo that mocks the database using a JSON file so that the tests run quicker since they don't have to spin up a DB.

ryasmi (dismissed):

davidpesce (approved):

PR #687, opened 2020-04-17
LMS 1689 replay failed statements

lzabo:
**Description**
- add a class to handle event processing named \logstore_xapi\log\process
This log processor require a parameter which is an array of log ids, second parameter is optional, input data came from failed or standard log table. 
- update store.php to use core function to get release data and to separate config data creating to a new function.

**PR Type**
- Enhancement

### DIFF
--- a/classes/log/process.php
+++ b/classes/log/process.php
@@ -1,0 +1,249 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


gordonmacqueen-lp (changes requested):
As per Michael's PR, we need to update the use of `$DB` to use `$repo` for database calls
https://github.com/xAPI-vle/moodle-logstore_xapi/pull/686#pullrequestreview-394604925

gordonmacqueen-lp (approved):

stephen-oh (approved):

PR #689, opened 2020-04-27
LMS-1797 - Generating info column for table

stephen-oh:
**Description**
- Generating info column for table using Response and Error Type

**Related Issues**
- N/A

**PR Type**
- Enhancement

### DIFF
--- a/lang/en/logstore_xapi.php
+++ b/lang/en/logstore_xapi.php
@@ -98,6 +98,11 @@ $string['privacy:metadata:logstore_xapi_log'] = 'xAPI holding table for cron pro
 $string['privacy:metadata:logstore_xapi_log:userid'] = 'User Id of xAPI holding table for cron processing';
 $string['privacy:metadata:logstore_xapi_failed_log'] = 'xAPI holding table for failed events';
 $string['privacy:metadata:logstore_xapi_failed_log:userid'] = 'User Id of xAPI holding table for failed events';
+$string['failedtosend'] = "The following statements have failed to be sent to the LDH.";
+$string['errorlogpage'] = "Error log page";
+$string['failurelog'] = "Failure log";
+$string['failedsubject'] = "XAPI Logstore: failed to send messages report";
+$string['norows'] = "No rows to report";
 
 // Info strings from xAPI errors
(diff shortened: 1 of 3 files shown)


lang/en/logstore_xapi.php
```diff
@@ -98,3 +98,12 @@
...
 $string['privacy:metadata:logstore_xapi_failed_log:userid'] = 'User Id of xAPI holding table for failed events';
+
+// Info strings from xAPI errors
+$string['networkerror'] = 'There was a network error sending the response: {$a}';
+$string['recipeerror'] = 'The LRS responded with a 400 error, this can be due to an issue with the recipe. The following message
+    was returned: "{$a}"';
+$string['autherror'] = 'The server is returning a 401 error: {$a}. Please ensure the endpoint, username and auth secret/password for
+    the xAPI is correct in the Logstore xAPI settings.';
+$string['unknownerror'] = 'Error code: "{$a->errortype}" Response: "{$a->response}"';
+$string['lrserror'] = 'There is a problem with the LRS. The LRS has responded with a 500 error. Response: {$a}';
```
gordonmacqueen-lp: JISC refer to an LRS as LDH - we should either change these strings, or add an additional parameter which picks up the value from another lang string

gordonmacqueen-lp (changes requested):

lang/en/logstore_xapi.php
mlynn-lp: We have to be careful with JISC specific changes and what the community of users would expect. Is LDH a JISC specific term?

gordonmacqueen-lp: I believe so @mlynn-lp

stephen-oh: That's why I left LRS in the key for the strings, by default these will be LDH (Learner Data Hub) but this can always be changed using the language pack. As this is the JISC branch, I don't see the issue really with sticking with LDH

gordonmacqueen-lp (approved):

mlynn-lp (approved):

lzabo (approved):

PR #692, opened 2020-04-28
LMS-2082: do not send duplicate emails

mlynn-lp:
**Description**
- Modify failed scheduled task to not send duplicate error messages.

**Related Issues**
- #2082

**PR Type**
- Enhancement

### DIFF
--- a/classes/task/sendfailednotifications_task.php
+++ b/classes/task/sendfailednotifications_task.php
@@ -124,10 +124,10 @@ class sendfailednotifications_task extends \core\task\scheduled_task {
     /**
      * Send email using email_to_user.
      *
-     * @param array failedrows an array of of failed rows from xapi_failed_log
-     * @param string msg email message
-     * @param string subject email subject
-     * @param string emailto email address for recipient
+     * @param array $failedrows an array of of failed rows from xapi_failed_log
+     * @param string $msg email message
+     * @param string $subject email subject
+     * @param string $emailto email address for recipient
      * 
      * @return int 1 = sent, 0 = not sent
      */
--- a/classes/task/sendfailednotifications_task.php
+++ b/classes/task/sendfailednotifications_task.php
@@ -124,7 +124,7 @@ class sendfailednotifications_task extends \core\task\scheduled_task {
     /**
      * Send email using email_to_user.
      *
-     * @param array $failedrows an array of failed rows from xapi_failed_log
+     * @param array $failedrows an array of failed rows from logstore_xapi_failed_log
      * @param string $msg email message
      * @param string $subject email subject
      * @param string $emailto email address for recipient
--- a/classes/task/sendfailednotifications_task.php
+++ b/classes/task/sendfailednotifications_task.php
@@ -126,7 +126,9 @@ class sendfailednotifications_task extends \core\task\scheduled_task {
      *
      * @return int 1 = sent, 0 = not sent
      */
-    private function sendmail($msg, $subject, $emailto) {
+    private function sendmail($failedrows, $msg, $subject, $emailto) {
+
+        global $DB;
 
         $user = new \stdClass();
         $user->id = self::DEFAULT_RECEIVER;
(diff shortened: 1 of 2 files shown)


classes/task/sendfailednotifications_task.php
```diff
@@ -126,7 +126,9 @@ private function get_failed_email_message($results) {
      *
```
gordonmacqueen-lp, line 126: missing param comments
what is failedrows, and what does it represent?

classes/task/sendfailednotifications_task.php
```diff
@@ -141,9 +143,30 @@ private function sendmail($msg, $subject, $emailto) {
...
-    
-        $messageid = email_to_user($user, $from, $subject, html_to_text($msg), $msg);
-        return $messageid;
+
+        // if any rows haven't been sent before then set flag to send
+        $rowstosend = false;
+        foreach ($failedrows as $row) {
+            $rv = $DB->get_records("logstore_xapi_notif_sent_log", array("failedlogid" => $row->id, "email" => $emailto));
+            if (empty($rv)) {
+                $rowstosend = true;
```
gordonmacqueen-lp, line 157: if you find a condition that sets `$rowstosend` to true, is there any need to continue iterating over the `$failedrows`?

gordonmacqueen-lp (changes requested):

classes/task/sendfailednotifications_task.php
mlynn-lp: Yes, I can exit the loop with a break; here.

classes/task/sendfailednotifications_task.php
mlynn-lp: I have added parameter comments. $failedrows is an array of of failed rows from xapi_failed_log

classes/task/sendfailednotifications_task.php
```diff
@@ -124,6 +124,11 @@ private function get_failed_email_message($results) {
     /**
      * Send email using email_to_user.
      *
+     * @param array failedrows an array of of failed rows from xapi_failed_log
```
gordonmacqueen-lp, line 127: typo: 'array of of failed'

classes/task/sendfailednotifications_task.php
gordonmacqueen-lp: typo: 'array **of of** failed...'

gordonmacqueen-lp: The dollar sign is required to denote param names
e.g. `* @param array $failedrows`
https://docs.phpdoc.org/latest/references/phpdoc/tags/param.html

mlynn-lp: I have added the $ sign in the comments.

classes/task/sendfailednotifications_task.php
```diff
@@ -141,9 +148,31 @@ private function sendmail($msg, $subject, $emailto) {
...
+            }
+        }
+
+        if ($rowstosend == true) {
+            $messageid = email_to_user($user, $from, $subject, html_to_text($msg), $msg);
+
+            // log that these notifications have been sent
+            $now = time();
+            $rows = array();
+            foreach ($failedrows as $row) {
```
stephen-oh, line 168: I'm wondering if it's necessary to loop these records twice?

classes/task/sendfailednotifications_task.php
```diff
@@ -141,9 +148,31 @@ private function sendmail($msg, $subject, $emailto) {
...
+
+        // if any rows haven't been sent before then set flag to send
+        $rowstosend = false;
+        foreach ($failedrows as $row) {
+            $rv = $DB->get_records("logstore_xapi_notif_sent_log", array("failedlogid" => $row->id, "email" => $emailto));
+            if (empty($rv)) {
+                $rowstosend = true;
+                break;
+            }
+        }
```
stephen-oh, line 160: I'm wondering if this is the best way to do this? Say we have 10 failed rows and only 1 of them appears in the sent log for that email address. That would mean the user will get 9 duplicate results in a new email with only 1 being a new result. Would it not make more sense to create a new array of records which have not been sent and use these in the email?

mlynn-lp: This means we would have to build a specific email for each user. The failed log captures the failed statements at that point in time regardless of what was sent to the user before. I think it could be confusing if they logged in and then saw the other failures which had been excluded from their most recent email just because they had received them previously.

classes/task/sendfailednotifications_task.php
mlynn-lp: I was thinking that too but what we are trying to find is if there is a need to send the email to this user because they haven't received any new alerts since the last time it was run. The first iteration is a bit shorter because it exits as soon as a new alert is detected.

classes/task/sendfailednotifications_task.php
gordonmacqueen-lp: `xapi_failed_log` should be `logstore_xapi_failed_log`

mlynn-lp: I have fixed that.

gordonmacqueen-lp (approved):

PR #694, opened 2020-04-30
LMS-1626 moodle-logstore-xapi: add ability to replay event individual…

lzabo:
LMS-1626 moodle-logstore-xapi: add ability to replay event individually from error log report.

**Description**
- add css to report page.
- add Ajax script to replay any number of event what a developer set.
- add javascript using amd to handled ajax script at the individual request.
- add some extra string to the lang file.

**Related Issues**
no related issue

**PR Type**
Enchancement


report.php
```diff
@@ -20,7 +20,9 @@
 require_once($CFG->dirroot . '/admin/tool/log/store/xapi/classes/form/reportfilter_form.php');
 
 define('XAPI_REPORT_STARTING_PAGE', 0);
-define('XAPI_REPORT_PERPAGE_DEFAULT', 30);
+define('XAPI_REPORT_PERPAGE_DEFAULT', 2);
```
lzabo: mistake :) I will set back.

classes/log/process.php
```diff
@@ -156,6 +170,20 @@ protected function get_event_ids(array $events) {
         }, $events);
     }
 
+    /**
+     * Get deletion event ids.
+     * Do not use array_map because it can throw an exception.
+     *
+     * @return array
+     */
+    private function get_delete_event_ids($loadedevents) {
```
lzabo: I will set it to protected.

report.php
```diff
@@ -134,12 +139,36 @@
...
     }
-    echo html_writer::start_tag('div', array('class'=>'no-overflow'));
+}
+
+// Create page view.
+$PAGE->set_context($systemcontext);
+$PAGE->set_url($baseurl);
+$PAGE->set_pagelayout('report');
+
+$PAGE->requires->js_call_amd('logstore_xapi/replayevents', 'init', [$eventids]);
```
lzabo, line 157: Issue here:
On the form I need a sum and perhaps all id of events but I don't have an idea how I can get it on the form easily.
thats why I pushed it in the js call. But it is too slow. Any better idea should be good for me.

stephen-oh: Could you use data attributes or perhaps hidden fields, depending on what you are needing to do with this?

lzabo: Because of the 'id updating' (when event processing run) I don't need the ids at all, only a sum I need.
I will remove $eventids parameter in the next PR. because this item is in connect with the multiple replaying.

lzabo: I am not sure how I can use data attributes in this case.
Could you give an explain?

report.php
```diff
@@ -134,12 +139,40 @@
...
+
+echo \html_writer::start_div('', ['id' => 'xapierrorlog']);
+echo \html_writer::start_div('', ['id' => 'xapierrorlog_form']);
+$mform->display();
+echo \html_writer::end_div();
+
+if (empty($results)) {
+    echo $OUTPUT->heading(get_string('noerrorsfound', 'logstore_xapi'));
+} else {
+    echo \html_writer::start_div('', ['class' => 'no-overflow', 'id' => 'xapierrorlog_data']);
```
lzabo: mistake, I will update it.

amd/src/replayevents.js
```diff
@@ -0,0 +1,382 @@
...
+             */
+            registerReplyEventListeners: function() {
+                var self = this;
+
+                $(SELECTORS.REPLAY_EVENTS).click(function(e) {
+                    e.stopPropagation();
+                    e.preventDefault();
+
+                    self.disableFormControls();
+
```
lzabo, line 112: perhaps also I should turn off the pagination (if it is on the page as well) during the request processing.

report.php
stephen-oh: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

Something like this is how I used to use PHP to generate data and store it on the page for reference by JS

lzabo: I see what you mean now.

amd/src/replayevents.js
mlynn-lp: I think the pagination would be necessary (and lower the server load) if a large number of rows were being processed.

mlynn-lp (commented):
styles.css needs an extra new line at the end of the file

report.php
```diff
@@ -22,17 +22,17 @@
 define('XAPI_REPORT_STARTING_PAGE', 0);
 define('XAPI_REPORT_PERPAGE_DEFAULT', 30);
 
+global $PAGE, $OUTPUT;
```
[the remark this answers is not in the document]
lzabo: I know just my editor wants it :)

classes/log/process.php
```diff
@@ -240,10 +266,18 @@ public function execute() {
...
+        // Store failed events.
+        $this->failedevents = $this->get_failed_events($loadedevents);
+        if(!empty($this->failedevents)) {
+            $this->store_failed_events($this->failedevents);
+        }
+
+        // Remove successful_events.
+        $this->successfulevents = $this->get_successful_events($loadedevents);
+        if(!empty($this->successfulevents)) {
+            //$this->record_successful_events($this->successfulevents);
```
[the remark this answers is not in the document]
lzabo: oh, I forgot to remove it good point.

report.php
```diff
@@ -134,12 +139,40 @@
...
-    echo html_writer::start_tag('div', array('class'=>'no-overflow'));
-    echo html_writer::table($table);
-    echo html_writer::end_tag('div');
-    echo $OUTPUT->paging_bar($count, $page, $perpage, $baseurl);
 }
 
+// Create page view.
+$PAGE->set_context($systemcontext);
+$PAGE->set_url($baseurl);
+$PAGE->set_pagelayout('report');
```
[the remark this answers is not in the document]
lzabo: Oh yes, I will remove my addition because it is not necessary. This file is yours we should use your solution. :)

gordonmacqueen-lp (commented):
This is really good @lzabo.

As I mentioned on the call, the idea would be that single and multiple replays would be handled in the same way
i.e. records are sent to the standard log to be replayed by the emit task

Having said that, I really like this solution but please bear in mind that the client may only want the simpler solution. Therefore, we should avoid spending much more time on this.

amd/src/replayevents.js
gordonmacqueen-lp: I think what @lzabo is saying is that the pagination should be disabled along with the rest of the controls when the user selects a single record for replay

stephen-oh: I was agreeing with @lzabo, @mlynn-lp's comment hadn't appeared for me

classes/log/process.php
```diff
@@ -156,6 +170,20 @@ protected function get_event_ids(array $events) {
         }, $events);
     }
 
+    /**
+     * Get deletion event ids.
+     * Do not use array_map because it can throw an exception.
+     *
```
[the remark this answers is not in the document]
lzabo: It was copied.

classes/form/reportfilter_form.php
```diff
@@ -42,6 +42,7 @@ public function definition() {
         $mform->addElement('date_selector', 'dateto', get_string('to'), ['optional' => true]);
 
         $this->add_action_buttons(false, get_string('search'));
+        $mform->addElement('submit', 'resendselected', get_string('resendevents', 'logstore_xapi', ['count' => 'loading']));
```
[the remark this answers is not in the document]
gordonmacqueen-lp: Just to clarify the requirement here incase it's not clear.

> The user can click a button labelled "Resend". If the error log has been filtered, only the statements associated with the filtered list are resent. Otherwise statements for all listed errors are re-sent.

I could be picking up the intention wrong from the variable name `resendselected`, but there's no need to select the individual records for replay with a checkbox. All records returned by the current filters form the set to be sent.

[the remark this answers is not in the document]
lzabo: Thanks for clarify this.

I didn't work to much on it so I will improve it.

PR #695, opened 2020-04-30
Lms 1626 replay single events

lzabo:
**Description**
add css to report page.
add Ajax script to move back any number of event what a developer set.
add javascript using amd to handle an individual request with ajax.
add some extra string to the lang file.

**Related Issues**
No related issue.

**PR Type**
Enhancement


amd/src/replayevents.js
```diff
@@ -0,0 +1,414 @@
...
+            },
+
+            /**
+             * Generate done icon.
+             */
+            generateDoneHTML : function(){
+                str.get_strings([
+                    {
+                        key: 'success',
+                        component: 'moodle'
```
lzabo, line 366: I will set it to come from the logstore_xapi module.

amd/src/replayevents.js
```diff
@@ -0,0 +1,414 @@
...
+            },
+
+            /**
+             * Generate load icon.
+             */
+            generateLoadHTML : function(){
+                str.get_strings([
+                    {
+                        key: 'loadinghelp',
+                        component: 'moodle'
```
lzabo, line 349: I will set it to come from the logstore_xapi module.

report.php
```diff
@@ -112,7 +105,9 @@
     }
     $table->head[] = get_string('info', 'logstore_xapi');
     $table->head[] = get_string('datetimegmt', 'logstore_xapi');
+    $table->head[] = '';
     $table->id = "report";
+    $replayevent = get_string('replayevent', 'logstore_xapi');
```
stephen-oh: Is this variable used?

lzabo: Nope, good catch. It was moved to the other place. I will remove it.

gordonmacqueen-lp (approved):

PR #696, opened 2020-05-01
Lms 2106 use send notification and error notification trigger

mlynn-lp:
**Description**
- Use send notification setting and error notification trigger limit

**Related Issues**
- #2106

**PR Type**
- Enhancement

### DIFF
--- a/classes/task/sendfailednotifications_task.php
+++ b/classes/task/sendfailednotifications_task.php
@@ -185,9 +185,21 @@ class sendfailednotifications_task extends \core\task\scheduled_task {
 
         echo "In send failed notifications task execute".PHP_EOL;
 
+        $enablesendingnotifications = get_config('logstore_xapi', 'enablesendingnotifications');
+        if (empty($enablesendingnotifications)) {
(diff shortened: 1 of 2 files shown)


classes/task/sendfailednotifications_task.php
```diff
@@ -185,12 +185,24 @@ public function execute() {
...
+        }
+
         $results = $this->get_failed_rows();
         if (count($results) == 0) {
             echo get_string('norows', 'logstore_xapi').PHP_EOL;
             return;
         }
 
+        $notificationtrigger = get_config('logstore_xapi', 'errornotificationtrigger');
+        if (count($results) < $notificationtrigger) {
```
stephen-oh, line 201: should this be less than or equal to? `<=`

mlynn-lp: I think < is correct. If $notificationtrigger is 10 it will trigger when the result set is 10 or more. Less than 10 it doesn't run.

gordonmacqueen-lp (approved):

PR #697, opened 2020-05-05
Lms 2115 add persistent

mlynn-lp:
**Description**
- Add logstoreid and type to logstore_standard_log_id and logstore_xapi_log and logstore_xapi_failed_log, populate in insert_event_entries().

**Related Issues**
- #2115

**PR Type**
- Enhancement

### DIFF
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -88,7 +88,7 @@ class store extends php_obj implements log_writer {
             'anonymous' => $event->anonymous
         ];
 
-        $sql = "SELECT MAX(ID) AS id
+        $sql = "SELECT MAX(id) AS id
                   FROM {logstore_standard_log}
                  WHERE eventname = :eventname
                    AND component = :component
@@ -100,18 +100,11 @@ class store extends php_obj implements log_writer {
                    AND userid = :userid
                    AND anonymous = :anonymous";
 
-        $rows = $DB->get_records_sql($sql, $sqlparams);
-
-        if (empty($rows)) {
+        $row = $DB->get_record_sql($sql, $sqlparams);
+        if (empty($row) || empty($row->id)) {
             return 0;
         }
-
-        $rowid = 0;
-        foreach ($rows as $row) {
-            $rowid = $row->id;
-            break;
-        }
-        return $rowid;
+        return $row->id;
     }
 
     /**
@@ -145,7 +138,7 @@ class store extends php_obj implements log_writer {
         foreach ($events as $event) {
             $eventid = $this->get_event_id($event);
             $event->logstorestandardlogid = $eventid;
-            $event->type = XAPI_TYPE_LIVE;
+            $event->type = XAPI_IMPORT_TYPE_LIVE;
         }
         return $events;
     }
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -65,7 +65,10 @@ class store extends php_obj implements log_writer {
     }
 
     /**
-     * get eventid from logstore_standard_log
+     * Get the eventid from logstore_standard_log.
+     * Return the persistent logstore_standard_log id from the database.
+     * Return value 0 means not found.
+     * 
      * @param event_base $event
      * @return int
      *
@@ -84,17 +87,19 @@ class store extends php_obj implements log_writer {
         $sqlparams['userid'] = $event->userid;
         $sqlparams['anonymous'] = $event->anonymous;
 
-        $rows = $DB->get_records_sql("SELECT MAX(ID) AS id
-                    FROM {logstore_standard_log}
-                    WHERE eventname = :eventname
-                    AND component = :component
-                    AND action = :action
-                    AND target = :target
-                    AND (objecttable = :objecttable OR objecttable IS NULL)
-                    AND (objectid = :objectid OR objectid IS NULL)
-                    AND timecreated = :timecreated
-                    AND userid = :userid
-                    AND anonymous = :anonymous", $sqlparams);
+        $sql = "SELECT MAX(ID) AS id
+                  FROM {logstore_standard_log}
+                 WHERE eventname = :eventname
+                   AND component = :component
+                   AND action = :action
+                   AND target = :target
+                   AND (objecttable = :objecttable OR objecttable IS NULL)
+                   AND (objectid = :objectid OR objectid IS NULL)
+                   AND timecreated = :timecreated
+                   AND userid = :userid
+                   AND anonymous = :anonymous";
+
+        $rows = $DB->get_records_sql($sql, $sqlparams);
 
         if (empty($rows)) {
             return 0;
@@ -114,6 +119,7 @@ class store extends php_obj implements log_writer {
      */
     protected function insert_event_entries(array $events) {
         global $DB;
+
         // If in background mode, just save them in the database.
         if ($this->get_config('backgroundmode', false)) {
             $events = $this->get_persistent_eventids($events);
@@ -129,16 +135,15 @@ class store extends php_obj implements log_writer {
 
     /**
      * Take rows from logstore_standard_log for the emit_task or failed_task
-     * and add in the logstore_standard_log_id and set the type.
+     * and add in the logstorestandardlogid and set the type.
      *
-     * Return the persistent logstore_standard_log id from the database.
      * @param array $events raw event data
      * @return array
      */
     private function get_persistent_eventids(array $events) {
         foreach ($events as $event) {
             $eventid = $this->get_event_id($event);
-            $event->logstore_standard_log_id = $eventid;
+            $event->logstorestandardlogid = $eventid;
             $event->type = XAPI_TYPE_LIVE;
         }
         return $events;
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -76,16 +76,17 @@ class store extends php_obj implements log_writer {
     protected function get_event_id($event) {
         global $DB;
 
-        $sqlparams = [];
-        $sqlparams['eventname'] = $event->eventname;
-        $sqlparams['component'] = $event->component;
-        $sqlparams['action'] = $event->action;
-        $sqlparams['target'] = $event->target;
-        $sqlparams['objecttable'] = $event->objecttable;
-        $sqlparams['objectid'] = $event->objectid;
-        $sqlparams['timecreated'] = $event->timecreated;
-        $sqlparams['userid'] = $event->userid;
-        $sqlparams['anonymous'] = $event->anonymous;
+        $sqlparams = [
+            'eventname' => $event->eventname, 
+            'component' => $event->component, 
+            'action' => $event->action, 
+            'target' => $event->target, 
+            'objecttable' => $event->objecttable, 
+            'objectid' => $event->objectid, 
+            'timecreated' => $event->timecreated, 
+            'userid' => $event->userid, 
+            'anonymous' => $event->anonymous
+        ];
 
         $sql = "SELECT MAX(ID) AS id
                   FROM {logstore_standard_log}
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -18,6 +18,7 @@ namespace logstore_xapi\log;
 defined('MOODLE_INTERNAL') || die();
 
 require_once(__DIR__ . '/../../src/autoload.php');
+require_once($CFG->dirroot . '/admin/tool/log/store/xapi/lib.php');
 
 use core_plugin_manager;
 use \tool_log\log\writer as log_writer;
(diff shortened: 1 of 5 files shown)


classes/log/store.php
```diff
@@ -63,15 +64,59 @@ protected function is_event_ignored(event_base $event) {
...
+        return $rowid;
+    }
+
     /**
      * Insert events in bulk to the database. Overrides helper_writer.
      * @param array $events raw event data
      */
     protected function insert_event_entries(array $events) {
         global $DB;
-
```
gordonmacqueen-lp, line 72: Please reinstate this empty line after the global declaration.

GH won't allow me to highlight the line, but there is a missing @ return comment for the above docblock

db/install.xml
```diff
@@ -27,6 +27,8 @@
         <FIELD NAME="origin" TYPE="char" LENGTH="10" NOTNULL="false" SEQUENCE="false" COMMENT="cli, cron, ws, etc."/>
         <FIELD NAME="ip" TYPE="char" LENGTH="45" NOTNULL="false" SEQUENCE="false" COMMENT="IP address"/>
         <FIELD NAME="realuserid" TYPE="int" LENGTH="10" NOTNULL="false" SEQUENCE="false" COMMENT="real user id when logged-in-as"/>
+        <FIELD NAME="logstoreid" TYPE="int" LENGTH="10" NOTNULL="false" SEQUENCE="false" COMMENT="logstore_standard_log_id"/>
```
gordonmacqueen-lp: I think the name should be `logstorestandardlogid` to avoid any ambiguity.

The `COMMENT` field contains underscores

db/install.xml
```diff
@@ -62,6 +64,8 @@
         <FIELD NAME="realuserid" TYPE="int" LENGTH="10" NOTNULL="false" SEQUENCE="false" COMMENT="real user id when logged-in-as"/>
         <FIELD NAME="errortype" TYPE="int" LENGTH="10" NOTNULL="false" SEQUENCE="false" COMMENT="The type of error (Statement send failure, Failure to authenticate with LDH, Network error, Malformed statement etc.)"/>
         <FIELD NAME="response" TYPE="text" NOTNULL="false" SEQUENCE="false" COMMENT="The response or response code received from the LDH"/>
+        <FIELD NAME="logstoreid" TYPE="int" LENGTH="10" NOTNULL="false" SEQUENCE="false" COMMENT="logstore_standard_log_id"/>
```
gordonmacqueen-lp: Same comments as per line 30 above

lib.php
```diff
@@ -18,6 +18,11 @@
 
 define('XAPI_REPORT_ID_ERROR', 0);
 define('XAPI_REPORT_ID_HISTORIC', 1);
+
+define('XAPI_TYPE_LIVE', 0);
+define('XAPI_TYPE_HISTORIC', 1);
+define('XAPI_TYPE_RECONCILE', 2);
+
```
gordonmacqueen-lp, line 26: what about an error type?

classes/log/store.php
```diff
@@ -63,15 +64,59 @@ protected function is_event_ignored(event_base $event) {
         return $isdisabledevent;
     }
 
+    /**
+     * get eventid from logstore_standard_log
+     * @param event_base $event
+     * @return int
```
gordonmacqueen-lp, line 73: should probably explain what is returned if/when the id cannot be found

classes/log/store.php
```diff
@@ -63,15 +64,59 @@ protected function is_event_ignored(event_base $event) {
...
+                    AND timecreated = :timecreated
+                    AND userid = :userid
+                    AND anonymous = :anonymous", $sqlparams);
+
+        if (empty($rows)) {
+            return 0;
+        }
+
+        $rowid = 0;
+        foreach ($rows as $row) {
```
gordonmacqueen-lp: is there any need to loop here?

classes/log/store.php
```diff
@@ -82,7 +127,25 @@ public function get_max_batch_size() {
...
+     * Take rows from logstore_standard_log for the emit_task or failed_task
+     * and add in the logstore_standard_log_id and set the type.
+     *
+     * Return the persistent logstore_standard_log id from the database.
+     * @param array $events raw event data
+     * @return array
+     */
+    private function get_persistent_eventids(array $events) {
+        foreach ($events as $event) {
+            $eventid = $this->get_event_id($event);
```
gordonmacqueen-lp, line 139: what happens if `get_event_id` returns 0?

classes/log/store.php
```diff
@@ -63,15 +64,59 @@ protected function is_event_ignored(event_base $event) {
...
+        global $DB;
+
+        $sqlparams = [];
+        $sqlparams['eventname'] = $event->eventname;
+        $sqlparams['component'] = $event->component;
+        $sqlparams['action'] = $event->action;
+        $sqlparams['target'] = $event->target;
+        $sqlparams['objecttable'] = $event->objecttable;
+        $sqlparams['objectid'] = $event->objectid;
+        $sqlparams['timecreated'] = $event->timecreated;
```
gordonmacqueen-lp: is `timecreated` a reliable param to check on here?

i.e. can we be sure the timestamp of the event is the same as that of the logstore_standard_log?

db/upgrade.php
```diff
@@ -64,6 +64,39 @@ function create_xapi_log_table($dbman, $tablename) {
     }
 }
 
+function create_logstoreid_and_type_columns($dbman) {
```
gordonmacqueen-lp: missing docblock

classes/log/store.php
```diff
@@ -63,15 +64,59 @@ protected function is_event_ignored(event_base $event) {
...
+                    FROM {logstore_standard_log}
+                    WHERE eventname = :eventname
+                    AND component = :component
+                    AND action = :action
+                    AND target = :target
+                    AND (objecttable = :objecttable OR objecttable IS NULL)
+                    AND (objectid = :objectid OR objectid IS NULL)
+                    AND timecreated = :timecreated
+                    AND userid = :userid
+                    AND anonymous = :anonymous", $sqlparams);
```
lzabo: Please use Moodle doc format styling when you create an SQL.

lzabo: What will happen if standard log is disabled?

classes/log/store.php
lzabo, line 72: I liked this extra line, could you add back?

lib.php
```diff
@@ -18,6 +18,11 @@
 
 define('XAPI_REPORT_ID_ERROR', 0);
 define('XAPI_REPORT_ID_HISTORIC', 1);
+
```
lzabo, line 21: remove this extra line or add a comment here, pls.

db/upgrade.php
```diff
@@ -64,6 +64,39 @@ function create_xapi_log_table($dbman, $tablename) {
...
+    $field = new xmldb_field('logstore_standard_log_id', XMLDB_TYPE_INTEGER, '10', null, null, null, null);
+    if (!$dbman->field_exists($table, $field)) {
+        $dbman->add_field($table, $field);
+    }
+
+    // Conditionally add field type to logstore_xapi_log.
+    $field = new xmldb_field('type', XMLDB_TYPE_INTEGER, '10', null, null, null, null);
+    if (!$dbman->field_exists($table, $field)) {
+        $dbman->add_field($table, $field);
+    }
```
gordonmacqueen-lp: this block is exactly the same as lines 87-97.

Wouldn't it better to rewrite as a `foreach` against an array containing the table names? 
If we change anything going forward, it means changing it once and eliminates the chance of a developer missing one of the tables/not duplicating it exactly.

db/install.xml
lzabo: The comment should be about "Reference to the id of standard log." (or something like that.)

db/install.xml
```diff
@@ -62,6 +64,8 @@
...
+        <FIELD NAME="type" TYPE="int" LENGTH="10" NOTNULL="false" SEQUENCE="false" COMMENT="xapi message type"/>
```
lzabo: The first letter should be capital in the comment or if you want to use as a name the correct should be xAPI because it is a instance type of API.

db/upgrade.php
```diff
@@ -64,6 +64,39 @@ function create_xapi_log_table($dbman, $tablename) {
...
+}
```
gordonmacqueen-lp, line 99: there's no indexes defined on the new columns

gordonmacqueen-lp (changes requested):

lzabo (changes requested):

classes/log/store.php
gordonmacqueen-lp: > What will happen if standard log is disabled?

Good point - I hadn't thought of that

classes/log/store.php
mlynn-lp: I have reinstated the missing line. The function does not return a value so no return value needs to be described.

db/install.xml
mlynn-lp: I have removed the underscores in the comment and changed this field to logstorestandardlogid and I also use that fieldname in upgrade.php

db/install.xml
mlynn-lp: I have removed the underscores in the comment and changed this field to logstorestandardlogid and I also use that fieldname in upgrade.php

lib.php
mlynn-lp: Errors go into the error table. This is more a type field to identify the import type.

classes/log/store.php
mlynn-lp: 0 could identify that the logstore entry is not found. I have added a comment to that effect.

classes/log/store.php
mlynn-lp: The returned row(s) contains an associative array keyd be the id of the first returned row. There might be no rows, so the quickest way is to iterate and return the first row.

classes/log/store.php
mlynn-lp: We insert the value 0 into the database. 0 means row not found. This prevents any problems with nulls.

classes/log/store.php
mlynn-lp: Yes the same data is passed to every logstore handler.

db/upgrade.php
mlynn-lp: I have added a docblock.

db/upgrade.php
lzabo: Or just put an extra parameter which  contains tablename in the function call.

mlynn-lp: I have put this in a function and called it with a table name parameter.

classes/log/store.php
mlynn-lp: Since replaying historical logs at the moment relies on data being inserted from logstore_standard_log I think if they want to use this feature they need to have it enabled.

classes/log/store.php
mlynn-lp: I have added it back.

db/install.xml
mlynn-lp: I have added this.

db/install.xml
mlynn-lp: I have corrected this.

classes/log/store.php
```diff
@@ -63,15 +64,59 @@ protected function is_event_ignored(event_base $event) {
...
+        $sqlparams['component'] = $event->component;
+        $sqlparams['action'] = $event->action;
+        $sqlparams['target'] = $event->target;
+        $sqlparams['objecttable'] = $event->objecttable;
+        $sqlparams['objectid'] = $event->objectid;
+        $sqlparams['timecreated'] = $event->timecreated;
+        $sqlparams['userid'] = $event->userid;
+        $sqlparams['anonymous'] = $event->anonymous;
+
+        $rows = $DB->get_records_sql("SELECT MAX(ID) AS id
```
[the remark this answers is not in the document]
mlynn-lp: I have fixed this. To prevent excessive tabbing I have written the SQL in a $sql variable.

classes/log/store.php
gordonmacqueen-lp: since you know `$rows` is not empty, can't you just do `$rows[0]->id`?

classes/log/store.php
gordonmacqueen-lp: My bad - I didn't realise the return was from the next function in the file
image

db/upgrade.php
mlynn-lp: I have added simple indexes to install.xml and upgrade.php. Type will probably be queried but logstorestandardlogid will probably not be queried.

lib.php
mlynn-lp: I have added a comment here to indicate the start of this section.

classes/log/store.php
mlynn-lp: In Moodle get_records_xxxx  always return arrays that have the value of the primary key id as indexes so $row[0]->id would not work. We get stuff like this:
```
array(1) {
  [1487]=>
  object(stdClass)#221 (1) {
    ["id"]=>
    string(4) "1487"
  }
}

```
The easiest way to iterate this is with a for loop.

gordonmacqueen-lp: Fair enough :)

db/upgrade.php
```diff
@@ -64,6 +64,43 @@ function create_xapi_log_table($dbman, $tablename) {
...
+ * 
+ * @param object $dbman
+ * @param string $tablename
+ *
+ */
+function add_logstorestandardlogid_type_to_table($dbman, $tablename) {
+    // Select table.
+    $table = new xmldb_table($tablename);
+
+    // Conditionally add field logstorestandardlogid to logstore_xapi_log.
```
gordonmacqueen-lp: Comment needs an update as it mentions a specific table.
The same applies on line 85

mlynn-lp: I have fixed this.

lib.php
gordonmacqueen-lp: After looking at this again, it looks like there's a bit of crossover between these three (`XAPI_TYPE_`) and the ones above (`XAPI_REPORT_ID_`).
Let's discuss at our catchup

classes/log/store.php
lzabo: array_keys() and after that keys[0] but your solution is the same and fine for me.

classes/log/store.php
```diff
@@ -63,6 +64,55 @@ protected function is_event_ignored(event_base $event) {
...
+        $sqlparams = [];
+        $sqlparams['eventname'] = $event->eventname;
+        $sqlparams['component'] = $event->component;
+        $sqlparams['action'] = $event->action;
+        $sqlparams['target'] = $event->target;
+        $sqlparams['objecttable'] = $event->objecttable;
+        $sqlparams['objectid'] = $event->objectid;
+        $sqlparams['timecreated'] = $event->timecreated;
+        $sqlparams['userid'] = $event->userid;
+        $sqlparams['anonymous'] = $event->anonymous;
```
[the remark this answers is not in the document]
mlynn-lp: Yes that's great.

lib.php
stephen-oh: @mlynn-lp can we change these to `XAPI_IMPORT_TYPE_` ?

classes/log/store.php
```diff
@@ -63,6 +64,56 @@ protected function is_event_ignored(event_base $event) {
...
+                   AND component = :component
+                   AND action = :action
+                   AND target = :target
+                   AND (objecttable = :objecttable OR objecttable IS NULL)
+                   AND (objectid = :objectid OR objectid IS NULL)
+                   AND timecreated = :timecreated
+                   AND userid = :userid
+                   AND anonymous = :anonymous";
+
+        $rows = $DB->get_records_sql($sql, $sqlparams);
```
stephen-oh: `MAX(ID)` should be `MAX(id)`?

If we only want 1 record then surely it would make more sense to use `get_record_sql'? Then you could remove the foreach below and replace with:
```
if (!empty($row)) {
   $rowid = $row->id;
}
```

lib.php
mlynn-lp: Yes I can do that.

classes/log/store.php
mlynn-lp: MAX(ID) still works but I'll change it to MAX(id). I was wondering if it would be possible that multiple rows could be returned by double logging in the same millisecond, so that's why I return the MAX(id). Originally it could potentially return multiple rows so I stuck with get_records_sql().

mlynn-lp: I have changed the code to use get_record_sql(); it saves a few lines of code.
Is it standard to use:
```
        if (empty($row) || empty($row->id)) {
            return 0;
        }
        return $row->id;
```
I tested it and it seems OK. No need to iterate the rows or use array_keys.

gordonmacqueen-lp (approved):

stephen-oh (approved):

PR #698, opened 2020-05-07
LMS-2179 moodle-logstore_xapi: try to avoid issue about get property …

lzabo:
try to avoid issue about get property on non-object.

**Description**
bugfix

**PR Type**
Fix

### DIFF
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -76,29 +76,68 @@ class store extends php_obj implements log_writer {
     protected function get_event_id($event) {
         global $DB;
 
-        $sqlparams = [
-            'eventname' => $event->eventname, 
-            'component' => $event->component, 
-            'action' => $event->action, 
-            'target' => $event->target, 
-            'objecttable' => $event->objecttable, 
-            'objectid' => $event->objectid, 
-            'timecreated' => $event->timecreated, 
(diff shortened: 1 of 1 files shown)


gordonmacqueen-lp (approved):

classes/log/store.php
```diff
@@ -260,4 +260,26 @@ protected function get_handler_config() {
...
     }
+
+    /**
+     * Reread or convert event to object.
+     *
+     * @param array $events Array of events
+     * @return array of objects of events.
+     */
+    protected function convert_array_to_objects($events) {
+        $return = array();
```
mlynn-lp, line 271: @lzabo convert_array_to_objects() doesn't seem to be called in this page.

classes/log/store.php
```diff
@@ -136,8 +175,7 @@ public function get_max_batch_size() {
      */
     private function get_persistent_eventids(array $events) {
         foreach ($events as $event) {
-            $eventid = $this->get_event_id($event);
-            $event->logstorestandardlogid = $eventid;
```
mlynn-lp, line 140: @lzabo This will fail unless if the $event is an array and not an object.

lzabo: I forget to add converter to the PR.

PR #699, opened 2020-05-07
Lms 2172 add database table of sent events

mlynn-lp:
**Description**
- Add table to support sent items.

**Related Issues**
- #2172

**PR Type**
- Enhancement


gordonmacqueen-lp:
Is this a complete copy of the structure of logstore standard log?
I don't think we need that level of detail, but keen to hear what others think 

For info, this is a table to support historical + reconciled records, so that we don't try to send them again if they've been successfully sent to the LRS previously

gordonmacqueen-lp (approved):

PR #703, opened 2020-05-12
LMS-2173: implement logging of sent events

mlynn-lp:
**Description**
- Populate table of sent events

**Related Issues**
- #2173

**PR Type**
- Enhancement

### DIFF
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -28,6 +28,7 @@ use \tool_log\helper\reader as helper_reader;
 use \tool_log\helper\buffered_writer as helper_writer;
 use \core\event\base as event_base;
 use \stdClass as php_obj;
+use stdClass;
 
 /**
  * This class processes events and enables them to be sent to a logstore.
@@ -168,6 +169,25 @@ class store extends php_obj implements log_writer {
     }
 
     /**
+     * Take event data from the emit_task
+     * and add to the sent log if it doesn't exist already.
+     *
+     * @param array $events raw event data
+     * @return array
+     */
+    private function add_event_to_sent_log($event) {
+        global $DB;
+        $row = $DB->get_record('logstore_xapi_sent_log', ['logstorestandardlogid' => $event->logstorestandardlogid]);
+        if (empty($row)) {
+            $newrow = new stdClass();
+            $newrow->logstorestandardlogid = $event->logstorestandardlogid;
+            $newrow->type = $event->type;
+            $newrow->timecreated = time();
+            $DB->insert_record('logstore_xapi_sent_log', $newrow);
+        }
+    }
+
+    /**
      * Take rows from logstore_standard_log for the emit_task or failed_task
      * and add in the logstorestandardlogid and set the type.
      *
@@ -178,6 +198,7 @@ class store extends php_obj implements log_writer {
         foreach ($events as $event) {
             $event->logstorestandardlogid = $this->get_event_id($event);
             $event->type = XAPI_IMPORT_TYPE_LIVE;
+            $this->add_event_to_sent_log($event);
         }
         return $events;
     }
--- a/classes/log/store.php
+++ b/classes/log/store.php
@@ -167,6 +167,16 @@ class store extends php_obj implements log_writer {
         return $this->get_config('maxbatchsize', 100);
     }
 
+    private function get_successful_events($events) {
+        $loadedevents = array_filter($events, function ($loadedevent) {
+            return $loadedevent['loaded'] === true;
+        });
+        $successfulevents = array_map(function ($loadedevent) {
+            return $loadedevent['event'];
+        }, $loadedevents);
+        return $successfulevents;
+    }
+
     /**
      * Take rows from logstore_standard_log for the emit_task or failed_task
      * and add in the logstorestandardlogid and set the type.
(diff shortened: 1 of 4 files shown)


classes/log/store.php
```diff
@@ -28,6 +28,7 @@
 use \tool_log\helper\buffered_writer as helper_writer;
 use \core\event\base as event_base;
 use \stdClass as php_obj;
+use stdClass;
```
stephen-oh: Is this a duplicate? as use stdClass is already set

classes/log/store.php
```diff
@@ -167,6 +168,25 @@ public function get_max_batch_size() {
...
+     * and add to the sent log if it doesn't exist already.
+     *
+     * @param array $events raw event data
+     * @return array
+     */
+    private function add_event_to_sent_log($event) {
+        global $DB;
+        $row = $DB->get_record('logstore_xapi_sent_log', ['logstorestandardlogid' => $event->logstorestandardlogid]);
+        if (empty($row)) {
+            $newrow = new stdClass();
```
stephen-oh: could this be `new php_obj();` ?

mlynn-lp: Hi Stephen,
Yes this could be php_obj();
The original code uses:
use \stdClass as php_obj;
Namespace aliases were added in PHP 5.6 and it is used in the extends clause in this class declaration but I don't think it adds clarity, so I used stdClass. Would you prefer me to use php_obj?

classes/log/store.php
mlynn-lp: Not a duplicate the first is aliased as php_obj

classes/log/store.php
lzabo: I understand your point, Michael but I don't think we should use more ways the same element so I agree with Stephen.
Otherwise we should delete the alias and update everywhere which is bigger task than we use this alias.

mlynn-lp: I will use the alias.

classes/log/store.php
```diff
@@ -167,6 +167,25 @@ public function get_max_batch_size() {
         return $this->get_config('maxbatchsize', 100);
     }
 
+    /**
+     * Take event data from the emit_task
+     * and add to the sent log if it doesn't exist already.
+     *
+     * @param array $events raw event data
+     * @return array
```
gordonmacqueen-lp: There is no return, so we should remove this comment

gordonmacqueen-lp (changes requested):

gordonmacqueen-lp:
Just to clarify @mlynn-lp, the requirement is to add to the log when we know the statement has been successful. I don't see anything in this PR covering this.

classes/log/store.php
mlynn-lp: I have removed this code and added i to the appropriate cron sub-task.

classes/log/store.php
mlynn-lp: I have done this, but moved the code to the appropriate cron sub-task.

mlynn-lp:
> Just to clarify @mlynn-lp, the requirement is to add to the log when we know the statement has been successful. I don't see anything in this PR covering this.

I have changed the code to filter on successful events in the cron sub-task. We will still have to do some processing in store.php however if the system is running in non-background mode. At the moment it doesn't filter on successful events in that code at all and I am wondering if the same filter functions can be applied in that code.

classes/task/emit_task.php
```diff
@@ -84,6 +85,36 @@ private function record_successful_events($events) {
         mtrace(count($this->get_successful_events($events)) . " " . get_string('successful_events', 'logstore_xapi'));
     }
 
+    /**
+     * Take event data from the current task
+     * and add to the sent log if it doesn't exist already.
+     *
+     * @param array $event raw event data
+     */
+    private function add_event_to_sent_log($event) {
```
gordonmacqueen-lp: Do you have any stats on the performance of this function?
From speaking with Ryan, it looks as though 1500 is the optimal batch size for sending a request to the LRS.

mlynn-lp: On the docker-moodle instance on my Mac with a 2.25GHz CPU with 16Gb RAM and SSD it took the following times to add 1500 rows on 3 separate tests:
1.8529641628265 seconds
1.689178943634 seconds
1.920863866806 seconds

classes/log/store.php
```diff
@@ -167,6 +167,16 @@ public function get_max_batch_size() {
         return $this->get_config('maxbatchsize', 100);
     }
 
+    private function get_successful_events($events) {
```
gordonmacqueen-lp: Has this been copied exactly from the emit task?
If so, should we move these functions out to the shared lib?

mlynn-lp: I'm always worried about cron tasks pulling in too much from lib.php
Being a private member function minimises the dependencies on lib.php
Would it be better to include this in lib.php?

gordonmacqueen-lp: I'd be keen to hear what the others think on this.
I'm thinking about duplication of code point of view, but can't comment from a moodle cron perspective

stephen-oh: I'm thinking we should avoid duplication of code to be honest. Perhaps we could use inheritance within these classes to avoid including the lib file?

lzabo: I don't like private function because it can not be inherited.
I don't think this function is necessary here.
I think All of these changes should be set in the emit task.

gordonmacqueen-lp:
This is ready to review again

classes/log/store.php
lzabo: This one is in the lib file now, isn't it?

classes/task/failed_task.php
```diff
@@ -124,7 +110,37 @@ private function store_failed_events($events) {
...
+    }
+
+    /**
+     * Take event data from the current task
+     * and add to the sent log if it doesn't exist already.
+     *
+     * @param array $event raw event data
+     */
+    private function add_event_to_sent_log($event) {
+        global $DB;
```
lzabo: Add an extra line after globals, pls.

mlynn-lp: I have done this.

classes/log/store.php
mlynn-lp: Yes it is in lib.php

gordonmacqueen-lp: This function should be removed from `classes/log/store.php`

classes/log/store.php
```diff
@@ -182,11 +192,43 @@ private function get_persistent_eventids(array $events) {
...
+        }
+    }
+
+    /**
+     * Take successful events and save each using add_event_to_sent_log.
+     *
+     * @param array $events raw events data
+     */
+    private function save_sent_events(array $events) {
+        $successfulevents = $this->get_successful_events($events);
```
gordonmacqueen-lp: This needs updated to refer to the version now in `lib.php`

gordonmacqueen-lp (changes requested):

gordonmacqueen-lp (approved):

PR #705, opened 2020-05-13
LMS-1692 moodle-logstore_xapi: add notifications.

lzabo:
**Description**
Add notifications

**Related Issues**
LSM-1692

**PR Type**
Enhancement

### DIFF
--- a/lang/en/logstore_xapi.php
+++ b/lang/en/logstore_xapi.php
@@ -123,3 +123,5 @@ $string['resendevents'] = 'Resend ({$a->count})';
 $string['replayevent'] = 'Replay event';
 $string['confirmresendeventsheader'] = 'Resend events';
 $string['confirmresendevents'] = 'You are about to send {$a->count} record(s) to the queue for reprocessing.<br>Do you wish to continue?';
+$string['resendevents:success'] = 'Events are set for reprocessing successfully.';
+$string['resendevents:failed'] = 'Events are set for reprocessing failed.';
--- a/lang/en/logstore_xapi.php
+++ b/lang/en/logstore_xapi.php
@@ -123,5 +123,5 @@ $string['resendevents'] = 'Resend ({$a->count})';
 $string['replayevent'] = 'Replay event';
 $string['confirmresendeventsheader'] = 'Resend events';
 $string['confirmresendevents'] = 'You are about to send {$a->count} record(s) to the queue for reprocessing.<br>Do you wish to continue?';
-$string['resendevents:success'] = 'Events are set for reprocessing successfully.';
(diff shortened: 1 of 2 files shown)


lang/en/logstore_xapi.php
```diff
@@ -123,3 +123,5 @@
 $string['replayevent'] = 'Replay event';
 $string['confirmresendeventsheader'] = 'Resend events';
 $string['confirmresendevents'] = 'You are about to send {$a->count} record(s) to the queue for reprocessing.<br>Do you wish to continue?';
+$string['resendevents:success'] = 'Events are set for reprocessing successfully.';
```
gordonmacqueen-lp: Change to 
"Events successfully sent for reprocessing"

lang/en/logstore_xapi.php
```diff
@@ -123,3 +123,5 @@
...
+$string['resendevents:failed'] = 'Events are set for reprocessing failed.';
```
gordonmacqueen-lp: Change to
"Events failed to be sent for reprocessing"

gordonmacqueen-lp (changes requested):

report.php
```diff
@@ -91,7 +94,11 @@
 
         if (!empty($eventids)) {
             $mover = new \logstore_xapi\log\moveback($eventids);
-            $mover->execute();  //LMS-1692 set notification
+            if ($mover->execute()) {
```
gordonmacqueen-lp, line 97: If there is an issue moving the records, do they all fail or can there be a mix between success/fail?

lzabo: I don't think there is any case when the mover will fail.

Only if the database handler throw error will the process failing. In this case we don't follow how many record were moved and how many were not.

lang/en/logstore_xapi.php
lzabo: great, thx.

report.php
gordonmacqueen-lp: ok - I just wondered if we knew a count we could display

lzabo: We know count but I think we should avoid to show it :)

lzabo: If we show the numbers that will give us more discussion like as:
- "If I set 10 element to replay and I confirm to update it why only 9 was move back?"
- "Because during our system waited for your confirmation, on the background the emit_task was run and resolved one of those."

gordonmacqueen-lp: Yeah I'm fine with that approach Lazslo.
I was just wondering if it was possible that some could fail but not others, and whether we had to consider that situation.
I'm happy with this, so there is nothing to do here.

gordonmacqueen-lp (approved):

PR #706, opened 2020-05-13
Lms 1701 extend report to list historical logstore records

stephen-oh:
**Description**
- Extending report to list historical logstore records and updating the filtering to be compatible with both

**Related Issues**
- N/A

**PR Type**
- Enhancement


classes/form/reportfilter_form.php
```diff
@@ -33,27 +33,44 @@ public function definition() {
...
+            $mform->addElement('select', 'errortype', get_string('errortype', 'logstore_xapi'), $errortypes);
+        }
+
+        $eventnameselect = $mform->addElement('select', 'eventnames', get_string('eventname', 'logstore_xapi'), $eventnames);
+        $eventnameselect->setMultiple(true);
+
+        if ($reportid == XAPI_REPORT_ID_ERROR) {
+            $mform->addElement('select', 'response', get_string('response', 'logstore_xapi'), $responses);
+        } elseif ($reportid == XAPI_REPORT_ID_HISTORIC) {
+            $mform->addElement('text', 'fullname', get_string('user', 'logstore_xapi'));
```
lzabo: We can use core's label here.

stephen-oh: Well spotted, I'll update this to `userfullname`

lang/en/logstore_xapi.php
```diff
@@ -104,6 +105,8 @@
 $string['norows'] = "No rows to report";
 $string['notificationsnotenabled'] = "Notifications not enabled";
 $string['notificationtriggerlimitnotreached'] = "Notification trigger limit not reached";
+$string['user'] = 'User';
```
lzabo, line 108: remove this one?

lzabo (approved):

lang/en/logstore_xapi.php
stephen-oh: I was thinking it was best to have a customisable string here as this field will display the username but in the spec it uses the title "user"

lang/en/logstore_xapi.php
```diff
@@ -89,6 +89,7 @@
 $string['info'] = 'Info';
 $string['datetimegmt'] = 'Date/Time (GMT)';
 $string['logstorexapierrorlog'] = 'Logstore xAPI Error Log';
+$string['logstorexapihistoriclog'] = 'Logstore xAPI Historic Log';
```
gordonmacqueen-lp, line 92: I'm not sure this is the correct name for it, but I'm not sure what it should be.
Does anyone else have any other ideas?

lzabo: Logstore xAPI - Error log's history

gordonmacqueen-lp: This one is not for errors. It is for Historical logstore events that the user wants to search for to send.

lzabo: I see, then:
Logstore xAPI log's history

mlynn-lp: Could it be "Logstore xAPI Historical Events Log" perhaps?

classes/form/reportfilter_form.php
```diff
@@ -33,27 +33,44 @@ public function definition() {
...
+        $reportid = $this->_customdata['reportid'];
         $eventnames = $this->_customdata['eventnames'];
-        $responses = $this->_customdata['responses'];
         $resend = $this->_customdata['resend'];
 
+        if ($reportid == XAPI_REPORT_ID_ERROR) {
+            $errortypes = $this->_customdata['errortypes'];
+            $responses = $this->_customdata['responses'];
+        } elseif ($reportid == XAPI_REPORT_ID_HISTORIC) {
+            $eventcontexts = $this->_customdata['eventcontexts'];
```
gordonmacqueen-lp, line 42: selecting an `eventcontext` in the form doesn't filter the results
image

classes/form/reportfilter_form.php
gordonmacqueen-lp: entering a username does not filter the results
image

classes/form/reportfilter_form.php
stephen-oh: I have fixed this if you wish to test

classes/form/reportfilter_form.php
stephen-oh: I have fixed this if you wish to test

gordonmacqueen-lp: It's working now 👍

classes/form/reportfilter_form.php
gordonmacqueen-lp: sorted 👍

gordonmacqueen-lp (approved):
Those items raised last night are working now.

There's a minor issue around lang strings when there are no entries to list but I'll raise a separate ticket for that
image

PR #709, opened 2020-05-19
LMS-1698 - fix relative path to config.php

gordonmacqueen-lp:
**Description**
- This fixes the issue when trying to execute the CLI script from anywhere other than the `admin/tool/log/store/xapi/cli` directory as per the Moodle Docs:
https://docs.moodle.org/dev/CLI_scripts#CLI_SCRIPT_constant

**Related Issues**
- LMS-1698

**PR Type**
- Fix

### DIFF
--- a/cli/testdataseeder.php
+++ b/cli/testdataseeder.php
@@ -1,28 +1,28 @@
 <?php
 define('CLI_SCRIPT', 1);
-include('../../../../../../config.php');
+require(__DIR__.'/../../../../../../config.php');
 
 function get_object() {
     $obj = new stdClass();
     $obj->eventname = '';
-    $obj->component = ''; 
-    $obj->action = ''; 
-    $obj->target = ''; 
+    $obj->component = '';
+    $obj->action = '';
+    $obj->target = '';
     $obj->objecttable = '';
-    $obj->objectid = ''; 
-    $obj->crud = ''; 
-    $obj->edulevel = ''; 
-    $obj->contextid = ''; 
-    $obj->contextlevel = ''; 
-    $obj->contextinstanceid = ''; 
-    $obj->userid = ''; 
-    $obj->courseid = ''; 
-    $obj->relateduserid = ''; 
-    $obj->anonymous = ''; 
-    $obj->other = ''; 
-    $obj->timecreated = time(); 
-    $obj->origin = ''; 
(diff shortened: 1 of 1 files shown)


gordonmacqueen-lp:
The real change is on line 3:
https://github.com/xAPI-vle/moodle-logstore_xapi/pull/709/files#diff-b0e2dfe02f0574366a9e85dca02bf850R3

My editor cleaned up the file and removed redundant whitespace.

stephen-oh (approved):

cli/testdataseeder.php
```diff
@@ -1,28 +1,28 @@
 <?php
 define('CLI_SCRIPT', 1);
-include('../../../../../../config.php');
+require(__DIR__.'/../../../../../../config.php');
```
lzabo, line 3: pls use require_once

cli/testdataseeder.php
```diff
@@ -1,28 +1,28 @@
 <?php
 define('CLI_SCRIPT', 1);
```
lzabo, line 2: Add file header things before define.

who created, this script is part of Moodle, copyright etc.

cli/testdataseeder.php
gordonmacqueen-lp: This came directly from the Moodle Docs. Do we need to change it?

>CLI_SCRIPT constant
All scripts must declare themselves as CLI scripts properly by setting the CLI_SCRIPT constant before loading the main config.php file.
>
>```
>define('CLI_SCRIPT', true);
>require(__DIR__.'/../../../config.php');
>```
>
>The path to the config.php must not use relative paths starting with ../. The best practise is to avoid dirname() function call and use the __DIR__ magic constant (this applies everywhere, not just for CLI scripts).

https://docs.moodle.org/dev/CLI_scripts#CLI_SCRIPT_constant

cli/testdataseeder.php
gordonmacqueen-lp: Good point.
This has already been merged, so we should address this when @mlynn-lp provides his next PR on this script.

cli/testdataseeder.php
lzabo: I see. I don't really like simple require because it can cause problems but at this case that's fine.

PR #713, opened 2020-05-20
Lms 1698 v3

mlynn-lp:
**Description**
- Create Users and support mod/assign submission and graded.

**Related Issues**
- #1698

**PR Type**
- Enhancement

### DIFF
--- a/cli/testdataseeder.php
+++ b/cli/testdataseeder.php
@@ -17,6 +17,7 @@
 
 define('CLI_SCRIPT', 1);
 require_once(__DIR__.'/../../../../../../config.php');
+require_once($CFG->libdir . '/testing/generator/lib.php');
 
 function get_object() {
     $obj = new stdClass();
@@ -129,6 +130,14 @@ function create_assignment_graded($table) {
     insert_row($table, $str);
 }
 
+/**
+ * Create a number of rows in the table.
+ * Add one row for each type, at least 10, pad the rest out with rows
+ * minus the number of message types.
+ *
+ * @param string $table tablename
+ * @param int $rows number of rows
+ */
 function create_test_data($table, $rows) {
     if ($rows < 10) {
         $rows = 10;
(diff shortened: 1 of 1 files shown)


cli/testdataseeder.php
```diff
@@ -115,15 +120,40 @@ function create_quiz_submitted($table) {
...
+ * minus the number of message types.
+ *
+ * @param string $table tablename
+ * @param int $rows number of rows
+ */
+function create_test_data($table, $rows) {
+    if ($rows < 10) {
+        $rows = 10;
+    }
+    $course_viewed = $rows - 8;
```
stephen-oh, line 153: What is the purpose of the $row's variable? Also if rows is < 10 then $course_viewed always equals 2? this is a bit confusing with what's in the doc block as well

cli/testdataseeder.php
```diff
@@ -132,8 +162,62 @@ function create_test_data($table) {
...
+    // TODO: create course we cannot restore a real course programmatically at the moment
+    // course should contain a quiz, forum and assignment
+    // the forum post and assignment submissions are an added complication
+    // assignment grade is related to the assignment submission
+}
+
+function create_data_set() {
+    // create n-rows in each table
+    // so we get one of each event and pad out the rest with course_viewed events
+    $rows = 10;
```
lzabo: should be defined in the top of the file?

cli/testdataseeder.php
```diff
@@ -132,8 +162,62 @@ function create_test_data($table) {
...
+
+    create_test_data("logstore_xapi_log", $rows);
+    create_test_data("logstore_xapi_failed_log", $rows);
+}
+
+create_standing_data();
+create_data_set();
 
 echo "Script complete".PHP_EOL;
```
lzabo: Yeah, I see, script is completed. I like it.

I would like to see something more interesting thing here:
Test data has been generated or Test data generating is finished.
Your English is much better than mine.

cli/testdataseeder.php
mlynn-lp: Yes that is a better idea.

cli/testdataseeder.php
mlynn-lp: I have changed the message to "Test data has been generated."

cli/testdataseeder.php
mlynn-lp: If the number of rows is less than 10 it is so that we get at least one message of each type and 2 course_viewed rows. It's just so the user doesn't specify 0 rows for example.

stephen-oh (approved):

gordonmacqueen-lp (approved):

PR #715, opened 2020-05-21
LMS-1649 - Bugfix - setting per page results from 30 to 40

stephen-oh:
**Description**
- According to spec the results per page should be 40 and not 30

**Related Issues**
- N/A

**PR Type**
- Fix


lzabo (approved):
Perhaps If we create a new setting with this parameter should be better.

gordonmacqueen-lp:
> Perhaps If we create a new setting with this parameter should be better.

It's a nice to have, but not in the requirements

gordonmacqueen-lp (approved):

PR #719, opened 2020-05-26
LMS-2285 - limit page access to admins only

gordonmacqueen-lp:
**Description**
- Adds the admin_externalpage_setup() call back into report.php to limit page access
- Move `$pagename` definition immediately after require_once

**Related Issues**
- LMS-2285

**PR Type**
- Fix

### DIFF
--- a/report.php
+++ b/report.php
@@ -21,6 +21,9 @@ require_once($CFG->dirroot . '/lib/adminlib.php');
 require_once($CFG->dirroot . '/admin/tool/log/store/xapi/lib.php');
 require_once($CFG->dirroot . '/admin/tool/log/store/xapi/classes/form/reportfilter_form.php');
 
+$pagename = 'logstorexapierrorlog';
+admin_externalpage_setup($pagename);
+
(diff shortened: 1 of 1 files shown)


stephen-oh (approved):

mlynn-lp (approved):

report.php
```diff
@@ -21,6 +21,9 @@
 require_once($CFG->dirroot . '/admin/tool/log/store/xapi/lib.php');
 require_once($CFG->dirroot . '/admin/tool/log/store/xapi/classes/form/reportfilter_form.php');
 
+$pagename = 'logstorexapierrorlog';
+admin_externalpage_setup($pagename);
```
lzabo, line 25: I don't think we have to add this because this set the page view to admin page which is not good for us.

only what we need:

>
     require_login(null, false);
    if (!has_capability('moodle/site:config', context_system::instance())) {
        print_error('accessdenied', 'admin');
    }

lzabo: If you set this a little bit down, around mform call you can use $systemcontext variable.

PR #721, opened 2020-05-27
LMS-2277 - Removing response from info column as this is output in fu…

stephen-oh:
…ll in the response column.

**Description**
- Removing response from info column as this is output in full in the response column

**Related Issues**
- N/A

**PR Type**
- Fix

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -201,28 +201,21 @@ function logstore_xapi_decode_response($response) {
  */
 function logstore_xapi_get_info_string($row) {
     if (!empty($row->errortype)) {
-        $response = '-';
-        if (isset($row->response)) {
-            $decode = logstore_xapi_decode_response($row->response);
-            if ($decode) {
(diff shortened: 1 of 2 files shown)


gordonmacqueen-lp (approved):

lib.php
```diff
@@ -201,28 +201,21 @@ function logstore_xapi_decode_response($response) {
...
     if (!empty($row->errortype)) {
-        $response = '-';
-        if (isset($row->response)) {
-            $decode = logstore_xapi_decode_response($row->response);
-            if ($decode) {
-                $response = $decode;
-            }
-        }
         switch ($row->errortype) {
             case 101:
```
lzabo, line 205: Could you define those variable?

lzabo: @Patches- mentioned about it would be solved in another PR.

lzabo (approved):

PR #725, opened 2020-06-01
Lms 2303 bugfix report.php missing contexts

stephen-oh:
**Description**
- Fixing error on report.php when a context no longer exists

**Related Issues**
- N/A

**PR Type**
- Fix


stephen-oh:
@gordonmacqueen-lp @mlynn-lp @lzabo It's worth mentioning that in my testing if a context no longer exists (such as a course) then it cannot be sent to the LRS/LDH as it won't transform as the context (course) doesn't exist.

gordonmacqueen-lp (approved):

lzabo (approved):

Issue #606, opened 2019-10-30
Transforming event having userid=0 creates exception


Callustian:
**Description**
- The transformer generates exceptions if the userid = 0 ( \core\event\course_viewed and \core\event\user_created). More generally the user ( guest? ) is not in the database, so every transformer that uses it fails.

**Version**
- https://github.com/xAPI-vle/moodle-logstore_xapi/releases/download/v4.2.5/xapi.zip

**Steps to reproduce the bug**
1. I am using Moodle, and I am unsure of how it is generating the event as guest, at the moment. This feature is not checked.

**Expected behaviour**
- events having userid=0 should be transformed because they exist.

**Actual behaviour**
- events having userid=0 create exception.

**Server information**
- Moodle+mysql.

**Client information**
- OS: linux
- Browser: chrome 78.0.3904.70 (Official Build) (64-bit)

Thank you.

davidpesce:
@Callustian - which version of the plugin are you using?

Callustian:
@davidpesce  Thank you for your reply.

I am using v4.2.5.

mlynn-lp:
@Callustian @davidpesce @gordonmacqueen-lp @Patches- @lzabo 
Suggested code change in these methods:
```
    // Get a valid user for guest.
    $guest = false;
    if ($event->userid == 0) {
        $guest = true;
        if (isset($CFG->siteguest)) {
            $event->userid = $CFG->siteguest;
        }
    }
    $user = $repo->read_record_by_id('user', $event->userid);

    // Reset userid so it is the same as in logstore_standard_log.
    if ($guest == true) {
        $event->userid = 0;
    }
```

Would it be possible to create a branch and put this in for code review in the following files:
course_viewed.php
user_created.php

mlynn-lp:
@ryansmith94 The exception is being thrown because Moodle stores userid 0 in the logstore when in fact the event was from the guest user.

I reset the userid to 0 so that the original userid is stored in xapi_logstore_log (same as it will be stored in logstore_standard_log) but for user retrieval purposes we know in some cases that the action was initiated by the guest user, so I retrieve the guest user details.

I think this may be a duplicate of issue #386.

davidpesce:
Closing due to inactivity as part of issue maintenance.

As a solo maintainer with limited bandwidth, I’m prioritizing issues that have recent activity and clear next steps.

If this is still relevant, please comment with an update:
[...]

PR #729, opened 2020-06-04
Issue 606: Transforming event having userid=0 creates exception fix.

mlynn-lp:
**Description**
- Transforming event having userid=0 creates exception fix.

**Related Issues**
- #606

**PR Type**
- Fix.

### DIFF
--- a/src/transformer/events/core/course_viewed.php
+++ b/src/transformer/events/core/course_viewed.php
@@ -25,7 +25,10 @@ function course_viewed(array $config, \stdClass $event) {
 
     $repo = $config['repo'];
 
-    // Get relateduserid or a valid userid for guest.
+    // Get the userid or a valid userid for guest.
+    if (!isset($event->userid)) {
+        $event->userid = 0;
+    }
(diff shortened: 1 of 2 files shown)


src/transformer/events/core/course_viewed.php
```diff
@@ -21,8 +21,13 @@
 use src\transformer\utils as utils;
 
 function course_viewed(array $config, \stdClass $event) {
+    global $CFG;
+
     $repo = $config['repo'];
-    $user = $repo->read_record_by_id('user', $event->userid);
+
+    // Get relateduserid or a valid userid for guest.
+    $userid = $event->userid == 0 && isset($CFG->siteguest) ? $CFG->siteguest : $event->userid;
```
gordonmacqueen-lp, line 32: @mlynn-lp 
Can the logic here be extracted out to a utility function?
It could take the config, event and an identifier for the field we are interested in (userid or relateduserid)

This way, it can be reused elsewhere too ([like here](https://github.com/xAPI-vle/moodle-logstore_xapi/pull/729/files#diff-5aced8c53c56680971e8f6104b187ab2R32)).

gordonmacqueen-lp (changes requested):

PR #730, opened 2020-06-10
Lms 2406 notification email content

stephen-oh:
**Description**
- Changing the email notifications to list only counts of each error type. Using mustache templates for email content. Logging only 1 row per user per notification email and changes to the notification threshold.

**Related Issues**
- LMS-2406

**PR Type**
- Enhancement

### DIFF
--- a/classes/output/renderer.php
+++ b/classes/output/renderer.php
@@ -14,12 +14,21 @@
 // You should have received a copy of the GNU General Public License
 // along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
 
+/**
+ * Renderer class for logstore_xapi
+ *
+ * @package    logstore_xapi
+ * @copyright  2020 Learning Pool Ltd <https://learningpool.com/>
+ * @author     Stephen O'Hara <stephen.ohara@learningpool.com>
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+namespace logstore_xapi\output;
+
 defined('MOODLE_INTERNAL') || die();
 
-$plugin = isset($plugin) && is_object($plugin) ? $plugin : new \stdClass();
-$plugin->component = 'logstore_xapi';
-$plugin->version = 2020061000;
+use plugin_renderer_base;
 
-$plugin->release = '';
-$plugin->requires = 2018051700;
-$plugin->maturity = MATURITY_STABLE;
+class renderer extends plugin_renderer_base {
+
+}
--- a/classes/task/sendfailednotifications_task.php
+++ b/classes/task/sendfailednotifications_task.php
@@ -31,8 +31,6 @@ class sendfailednotifications_task extends \core\task\scheduled_task {
      * Constants
      * Repurpose email_to_user() to send for users with just email addresses.
      */
-    const DEFAULT_RECEIVER = -99;
-    const DEFAULT_RECEIVER_NAME = "";
     const DEFAULT_SENDER = -99;
     const DEFAULT_SENDER_NAME = "";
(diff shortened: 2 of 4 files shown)


classes/output/renderer.php
```diff
@@ -0,0 +1,34 @@
...
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+namespace logstore_xapi\output;
+
+defined('MOODLE_INTERNAL') || die();
+
+use plugin_renderer_base;
+
+class renderer extends plugin_renderer_base {
```
lzabo, line 32: the name should be logstore_xapi_renderer

stephen-oh: I'm using namespace, see line 26

lzabo: https://docs.moodle.org/dev/Overriding_a_renderer

> There is also a routine way to name renderers much like the process for naming a theme's overriding renderer.
> 
> It should always be made up of the following three components joined by underscores:
> 
> component This is the component the renderer belongs too, e.g. block, mod, or core.
> name This is the name of what ever the component is, e.g. forum, workshop, or wiki.
> renderer They should always be completed by the word renderer.

stephen-oh: https://docs.moodle.org/dev/Overriding_a_renderer#Namespaces

> Renderers can also be used with code that uses namespaces. When creating or overriding new renderers with namespaces, be sure to use "output" as the second level namespace to comply with the rules for namespaces.

lzabo: Yes, thats true. 

Renderers can also be used with code that uses namespaces.

lzabo: So you need to use namespace and follow the naming convention.

gordonmacqueen-lp (approved):

mlynn-lp (approved):

classes/task/sendfailednotifications_task.php
```diff
@@ -47,93 +45,70 @@ public function get_name() {
     }
 
     /**
-     * Get failed rows.
+     * Get the last id of the new notifications to send, if there are no notifications to send then return false.
      *
-     * @return array
+     * @return array|bool
```
gordonmacqueen-lp: I think this should now be `int|bool`

classes/task/sendfailednotifications_task.php
```diff
@@ -47,93 +45,70 @@ public function get_name() {
...
     }
 
     /**
      * Send email using email_to_user.
      *
-     * @param array $failedrows an array of failed rows from logstore_xapi_failed_log
      * @param string $message email message
      * @param string $subject email subject
      * @param object $user user to receive email
+     * @param $lastfailedid
```
gordonmacqueen-lp: int?

gordonmacqueen-lp (changes requested):

classes/task/sendfailednotifications_task.php
```diff
@@ -47,93 +45,70 @@ public function get_name() {
...
+        $lastnotification = $DB->get_record_sql('SELECT MAX(failedlogid) AS id FROM {logstore_xapi_notif_sent_log}');
+
+        if ($lastnotification) {
+            $lastnotificationid = $lastnotification->id;
+        }
+
+        $sql = 'SELECT MAX(id) AS id
+                  FROM {logstore_xapi_failed_log}
+                 WHERE id > :lastsentid
+                HAVING COUNT(id) > :threshold';
```
gordonmacqueen-lp: `>=`

gordonmacqueen-lp (approved):

PR #737, opened 2020-06-18
Lms 2481 bugfix source table may be incorrect

stephen-oh:
**Description**
- LMS-2481 - Bugfix for failed task. Failed task now only resends from the XAPI_LOG table using the correct type. Added Failed type to moveback class

**Related Issues**
- N/A

**PR Type**
- Fix


classes/task/failed_task.php
```diff
@@ -78,21 +39,12 @@ public function execute() {
...
+        $loadedevents = $store->process_events($extractedevents, XAPI_IMPORT_TYPE_FAILED);
 
-        // Re-run as normal.
-        $extractedevents = logstore_xapi_extract_events($batchsize);
-        $loadedevents = $store->process_events($extractedevents);
- 
         logstore_xapi_store_failed_events($loadedevents);
         logstore_xapi_record_successful_events($loadedevents);
         logstore_xapi_save_sent_events($loadedevents);
         logstore_xapi_delete_processed_events($loadedevents);
```
lzabo: My suggestion:
We should those functions set back to emit_task as a protected functions.
All other task should be a child of emit_task.

Result:
- Smaller lib file
- Cleaner code

lzabo: Also if we add two parameter to the emit_task:
- import_type
- batch size
And set those in the __construct functions in all tasks well, we don't need to add any extra function to the other tasks.

gordonmacqueen-lp (commented):
It's mostly there, but none of the tasks complete now and exit with an error complaining about a function it can no longer find from the lib.

```
!!! Exception - Call to undefined function logstore_xapi\log\logstore_xapi_get_successful_events() !!!
!!
Error code: generalexceptionmessage !!
!! Stack trace: * line 200 of /admin/tool/log/store/xapi/classes/log/store.php: Error thrown
* line 220 of /admin/tool/log/store/xapi/classes/log/store.php: call to logstore_xapi\log\store-&gt;save_sent_events()
* line 43 of /admin/tool/log/store/xapi/classes/task/historical_task.php: call to logstore_xapi\log\store-&gt;process_events()
* line 157 of /admin/tool/task/cli/schedule_task.php: call to logstore_xapi\task\historical_task-&gt;execute()
 !!
```

This function then makes a call to `logstore_xapi_add_event_to_sent_log` which is no longer in the lib either.

stephen-oh:
This was merged by @gordonmacqueen-lp in another PR

PR #740, opened 2020-06-24
LMS-2580: fix XAPI_IMPORT_TYPE_LIVE (0) type check

mlynn-lp:
**Description**
- Check if XAPI_IMPORT_TYPE_LIVE (0) is being sent. Before this change if 0 was passed $conditions would not be set. In the original code before refactoring this method was used in the process() method with optional params. Since param 2 is optional, param 3 must also be optional.

**Related Issues**
- #2580

**PR Type**
- Fix.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -282,7 +282,7 @@ function logstore_xapi_extract_events($limitnum = 0, $log = XAPI_REPORT_SOURCE_L
     global $DB;
 
     $conditions = null;
-    if ($type) {
+    if (!empty($type) || $type == XAPI_IMPORT_TYPE_LIVE) {
         $conditions = array("type" => $type);
     }
     $sort = '';
--- a/lib.php
+++ b/lib.php
@@ -278,7 +278,7 @@ function logstore_xapi_add_event_to_sent_log($event) {
  * @param int $type event type
  * @return array
  */
-function logstore_xapi_extract_events($limitnum = 0, $log, $type) {
+function logstore_xapi_extract_events($limitnum, $log, $type) {
     global $DB;
 
     $conditions = array("type" => $type);
--- a/lib.php
+++ b/lib.php
@@ -278,13 +278,10 @@ function logstore_xapi_add_event_to_sent_log($event) {
  * @param int $type event type
  * @return array
  */
-function logstore_xapi_extract_events($limitnum = 0, $log = XAPI_REPORT_SOURCE_LOG, $type = null) {
+function logstore_xapi_extract_events($limitnum = 0, $log, $type) {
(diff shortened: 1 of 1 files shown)


lib.php
```diff
@@ -282,7 +282,7 @@ function logstore_xapi_extract_events($limitnum = 0, $log = XAPI_REPORT_SOURCE_L
     global $DB;
 
     $conditions = null;
-    if ($type) {
+    if (!empty($type) || $type == XAPI_IMPORT_TYPE_LIVE) {
```
stephen-oh: Perhaps we want `===` here?

mlynn-lp: === is useful to match value and type but is type important when using define to declare a constant?

gordonmacqueen-lp: Why don't we just get rid of the optional'ness of the (2nd and 3rd?) params if they are always provided?
That way, you can get rid of the if statement entirely

gordonmacqueen-lp: I've just checked, and everywhere the function is used all three params are provided
```
gordonmacqueen@LPOOL55- store/xapi (master-jisc) » grep -r logstore_xapi_extract_events *                                                            141 ↵
classes/task/failed_task.php:        $extractedevents = logstore_xapi_extract_events($batchsize, XAPI_REPORT_SOURCE_LOG, XAPI_IMPORT_TYPE_FAILED);
classes/task/historical_task.php:        $extractedevents = logstore_xapi_extract_events($batchsize, XAPI_REPORT_SOURCE_LOG, XAPI_IMPORT_TYPE_HISTORIC);
classes/task/emit_task.php:        $extractedevents = logstore_xapi_extract_events($batchsize, XAPI_REPORT_SOURCE_LOG, XAPI_IMPORT_TYPE_LIVE);
lib.php:function logstore_xapi_extract_events($limitnum = 0, $log = XAPI_REPORT_SOURCE_LOG, $type = null)
```

lib.php
```diff
@@ -278,13 +278,10 @@ function logstore_xapi_add_event_to_sent_log($event) {
  * @param int $type event type
  * @return array
  */
-function logstore_xapi_extract_events($limitnum = 0, $log = XAPI_REPORT_SOURCE_LOG, $type = null) {
+function logstore_xapi_extract_events($limitnum = 0, $log, $type) {
```
gordonmacqueen-lp: I think you'll also need to get rid of the default value for `$limitnum` 

https://www.php.net/manual/en/functions.arguments.php
see: Examples 4, 5 + 6

gordonmacqueen-lp (approved):

PR #741, opened 2020-06-25
LMS-1698: add historical items to logstore_xapi_log with type XAPI_IM…

mlynn-lp:
**Description**
- Add historical items to logstore_xapi_log with type XAPI_IMPORT_TYPE_HISTORIC.

**Related Issues**
- #1698

**PR Type**
- Enhancement

### DIFF
--- a/cli/testdataseeder.php
+++ b/cli/testdataseeder.php
@@ -63,11 +63,6 @@ function insert_row($table, $rowcsv) {
     $strarr = explode(",", $rowcsv);
     $type = logstore_xapi_get_type_from_table($table);
 
-    if ($table == XAPI_REPORT_SOURCE_HISTORICAL) {
-        $table = XAPI_REPORT_SOURCE_LOG;
-        $type = XAPI_IMPORT_TYPE_HISTORIC;
-    }
-
     $n = 0;
     foreach ($obj as $key => $value) {
         $clean = clean_string($strarr[$n]);
--- a/cli/testdataseeder.php
+++ b/cli/testdataseeder.php
@@ -63,7 +63,7 @@ function insert_row($table, $rowcsv) {
     $strarr = explode(",", $rowcsv);
     $type = logstore_xapi_get_type_from_table($table);
 
-    if ($table == XAPI_IMPORT_TYPE_HISTORIC) {
+    if ($table == XAPI_REPORT_SOURCE_HISTORICAL) {
         $table = XAPI_REPORT_SOURCE_LOG;
         $type = XAPI_IMPORT_TYPE_HISTORIC;
     }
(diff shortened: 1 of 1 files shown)


cli/testdataseeder.php
```diff
@@ -219,6 +225,7 @@ function create_standing_data() {
 function create_data_set() {
     create_test_data(XAPI_REPORT_SOURCE_LOG, ROWS);
     create_test_data(XAPI_REPORT_SOURCE_FAILED, ROWS);
+    create_test_data(XAPI_IMPORT_TYPE_HISTORIC, ROWS);
```
gordonmacqueen-lp: this should be `XAPI_REPORT_SOURCE_HISTORICAL`

cli/testdataseeder.php
```diff
@@ -63,6 +63,11 @@ function insert_row($table, $rowcsv) {
     $strarr = explode(",", $rowcsv);
     $type = logstore_xapi_get_type_from_table($table);
 
+    if ($table == XAPI_IMPORT_TYPE_HISTORIC) {
+        $table = XAPI_REPORT_SOURCE_LOG;
+        $type = XAPI_IMPORT_TYPE_HISTORIC;
+    }
+
```
gordonmacqueen-lp: The line `$table = XAPI_REPORT_SOURCE_LOG;` is incorrect as it should be pointing at the standard log table, however If you send the table as `XAPI_REPORT_SOURCE_HISTORICAL` (as in the other comment) there is no need for any of this at all.

I added the `logstore_xapi_get_type_from_table` function which should return `XAPI_IMPORT_TYPE_HISTORIC` when supplying `XAPI_REPORT_SOURCE_HISTORICAL` as the `$table`

gordonmacqueen-lp (changes requested):

cli/testdataseeder.php
mlynn-lp: This table name swap was just to avoid further changes, but I have corrected this to use a table name rather than type constant in the caller.

cli/testdataseeder.php
mlynn-lp: I have changed this. The destination table and type is swapped if that had been passed.

cli/testdataseeder.php
gordonmacqueen-lp: I think this is still incorrect @mlynn-lp 
The intention is to add rows to `mdl_logstore_standard_log` to allow QA to select historical entries to be sent on to `mdl_logstore_xapi_log`. As it stands, this will write to `mdl_logstore_xapi_log`

gordonmacqueen-lp (approved):

PR #743, opened 2020-06-26
LMS-2609: LRS endpoint link on email not working should be LMS instan…

mlynn-lp:
**Description**
- LRS endpoint link on email not working. Should be LMS URL.

**Related Issues**
- #2609

**PR Type**
- Fix.

### DIFF
--- a/lang/en/logstore_xapi.php
+++ b/lang/en/logstore_xapi.php
@@ -128,3 +128,4 @@ $string['confirmresendevents'] = 'You are about to send {$a->count} record(s) to
 $string['confirmsendevents'] = 'You are about to send {$a->count} record(s) to the queue for reprocessing.<br>Do you wish to continue?';
 $string['resendevents:success'] = 'Events successfully sent for reprocessing.';
 $string['resendevents:failed'] = 'Events failed to be sent for reprocessing.';
+$string['lmsinstance'] = 'LMS instance on which the error(s) occurred';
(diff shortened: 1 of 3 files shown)


lang/en/logstore_xapi.php
```diff
@@ -128,3 +128,4 @@
 $string['confirmsendevents'] = 'You are about to send {$a->count} record(s) to the queue for reprocessing.<br>Do you wish to continue?';
 $string['resendevents:success'] = 'Events successfully sent for reprocessing.';
 $string['resendevents:failed'] = 'Events failed to be sent for reprocessing.';
+$string['lmsinstance'] = 'LMS instance on which the error(s) occurred';
```
gordonmacqueen-lp, line 131: I know that's what the link is, but it doesn't read too well in an email

> The following statements have failed to be sent to the LDH.
>
>
> [LMS instance on which the error(s) occurred](url)
>
> [Error log page](url)
>
> ...table of errors...

mlynn-lp: Yes, should we change the layout? In the very first version I was going to put the links above the header.

gordonmacqueen-lp: Yeah it just needs to flow a bit better than it does just now.

How about following the order as defined in the spec?
> Preamble text explaining that there was an error sending statements to the LDH
The details of the error(s), including the time of the error(s), the response from the LDH, and the type of the error(s)
The name/url of the LMS instance on which the error(s) occurred
A link to the error log page on the LMS

gordonmacqueen-lp (approved):

PR #745, opened 2020-06-30
LMS-2657 - send username 

gordonmacqueen-lp:
**Description**
- resolves issue where username not being sent in `actor.account.name` when 'Identify users by id' is checked and 'Identify users by email' is not checked.

**Related Issues**
- #733

**PR Type**
- Fix

PR #747, opened 2020-07-03
Lms 2589 reconcile mod assign assignment submitted

lzabo:
**Description**
- add get_verb function
- set verb to completed at the assignment_submitted when JISC data is set.

**Related Issues**
- LMS-2589

**PR Type**
- Enhancement

### DIFF
--- a/src/transformer/events/mod_assign/assignment_submitted.php
+++ b/src/transformer/events/mod_assign/assignment_submitted.php
@@ -28,14 +28,15 @@ function assignment_submitted(array $config, \stdClass $event) {
     $assignment = $repo->read_record_by_id('assign', $assignmentsubmission->assignment);
     $lang = utils\get_course_lang($course);
 
+    $verb = utils\get_verb($config, $lang, 'submit');
+
+    if (utils\is_enabled_config($config, 'send_jisc_data')) {
+        $verb = utils\get_verb($config, $lang, 'completed');
+    }
+
     return [[
         'actor' => utils\get_user($config, $user),
-        'verb' => [
-            'id' => 'http://activitystrea.ms/schema/1.0/submit',
-            'display' => [
-                $lang => 'submitted'
-            ],
-        ],
+        'verb' => $verb,
         'object' => utils\get_activity\course_assignment($config, $event->contextinstanceid, $assignment->name, $lang),
         'timestamp' => utils\get_event_timestamp($event),
         'context' => [
--- a/src/transformer/events/mod_assign/assignment_submitted.php
+++ b/src/transformer/events/mod_assign/assignment_submitted.php
@@ -28,10 +28,10 @@ function assignment_submitted(array $config, \stdClass $event) {
     $assignment = $repo->read_record_by_id('assign', $assignmentsubmission->assignment);
     $lang = utils\get_course_lang($course);
 
-    $verb = utils\get_verb($config, $lang, 'submitted');
+    $verb = utils\get_verb('submitted', $config, $lang);
 
     if (utils\is_enabled_config($config, 'send_jisc_data')) {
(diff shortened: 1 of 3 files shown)


src/transformer/utils/get_user.php
```diff
@@ -19,6 +19,11 @@
 
 function get_user(array $config, \stdClass $user) {
     $fullname = get_full_name($user);
+
+    if (is_enabled_config($config, 'send_jisc_data')) {
+        $fullname = $user->username;
+    }
+
```
gordonmacqueen-lp: as discussed, this is not required.
It is covered by #745 / #746

gordonmacqueen-lp (changes requested):

src/transformer/events/mod_assign/assignment_submitted.php
```diff
@@ -28,14 +28,15 @@ function assignment_submitted(array $config, \stdClass $event) {
     $assignment = $repo->read_record_by_id('assign', $assignmentsubmission->assignment);
     $lang = utils\get_course_lang($course);
 
+    $verb = utils\get_verb($config, $lang, 'submit');
```
gordonmacqueen-lp: this will fail
In the `get_verb` function the case is `submitted`

src/transformer/utils/get_user.php
lzabo: I will remove it.

src/transformer/events/mod_assign/assignment_submitted.php
lzabo: good spot

gordonmacqueen-lp (commented):
There are other events containing the 'submitted' and 'completed' verbs 
e.g. courses, attempts, modules etc

Therefore, I think we should change the `get_verb` util to be `get_assignment_verb` and perform the 'send_jisc_check' inside the util

I think this approach is easier to read, maintain and avoids potential clashes further down the line.
It also follows the pattern already in place for [get_scorm_verb](https://github.com/xAPI-vle/moodle-logstore_xapi/blob/master/src/transformer/utils/get_scorm_verb.php)

lzabo:
> There are other events containing the 'submitted' and 'completed' verbs
> e.g. courses, attempts, modules etc
> 
> Therefore, I think we should change the `get_verb` util to be `get_assignment_verb` and perform the 'send_jisc_check' inside the util
> 
> I think this approach is easier to read, maintain and avoids potential clashes further down the line.
> It also follows the pattern already in place for [get_scorm_verb](https://github.com/xAPI-vle/moodle-logstore_xapi/blob/master/src/transformer/utils/get_scorm_verb.php)

I don’t agree.

I guess we need a function which contains the correct verbs and also that will be great if we add these verb settings to the other functions as well.

pros:
All verb will be in one place.
Easy to follow standards.
Developer friendly version, It helps to avoid duplicated elements and to avoid duplicated code and to avoid spelling mistakes.

cons:
.

The get_scorm_verb function comes back with different statuses which is good but in my version it should call my get_verb function three times, instead of they add time to time verbs, (for example completed verb).

gordonmacqueen-lp:
> I don’t agree.
> 
> I guess we need a function which contains the correct verbs and also that will be great if we add these verb settings to the other functions as well.
> 
> pros:
> All verb will be in one place.
> Easy to follow standards.
> Developer friendly version, It helps to avoid duplicated elements and to avoid duplicated code and to avoid spelling mistakes.
> 
> cons:
> .
> 
> The get_scorm_verb function comes back with different statuses which is good but in my version it should call my get_verb function three times, instead of they add time to time verbs, (for example completed verb).

OK - let's go with your suggested approach. 
I had misunderstood that the verbs are reused between the different event types

PR #748, opened 2020-07-05
LMS-2668 reconcile loggedin event

gordonmacqueen-lp:
**Description**
- amends verb id when 'send_jisc_data' is enabled

**Related Issues**
- #LMS-2668

**PR Type**
- Enhancement

### DIFF
--- a/src/transformer/utils/get_verb.php
+++ b/src/transformer/utils/get_verb.php
@@ -18,6 +18,8 @@ namespace src\transformer\utils;
 
 defined('MOODLE_INTERNAL') || die();
 
+use src\transformer\utils as utils;
+
 /**
  * Return the requested verb with details.
  *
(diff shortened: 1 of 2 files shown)


src/transformer/utils/get_verb.php
```diff
@@ -18,6 +18,8 @@
 
 defined('MOODLE_INTERNAL') || die();
 
+use src\transformer\utils as utils;
```
lzabo, line 21: This namespace is the same as where our function have, so I am not sure why we need this to add.

src/transformer/utils/get_verb.php
```diff
@@ -49,6 +51,20 @@ function get_verb($verb, array $config, $lang) {
...
+        case 'loggedin':
+            $output = [
+                'id' => 'https://brindlewaye.com/xAPITerms/verbs/loggedin/',
+                'display' => [
+                    $lang = 'logged into'
+                ]
+            ];
+
+            // JISC specific verb id
+            if (utils\is_enabled_config($config, 'send_jisc_data')) {
```
lzabo, line 63: we can use a simple function call here only because that one is in the same namespace.

is_enabled_config($config, 'send_jisc_data')

src/transformer/utils/get_verb.php
gordonmacqueen-lp: I added it to stay consistent with the style used in the rest of the plugin.
https://github.com/xAPI-vle/moodle-logstore_xapi/blob/master/src/transformer/utils/get_multichoice_definition.php

lzabo (approved):

lzabo (approved):

stephen-oh (approved):

PR #750, opened 2020-07-06
Lms 2589 reconcile mod assign assignment submitted

lzabo:
**Description**
Reconcile mod assign assignment submitted.
- Populate context extention for vle_mod_id
- Capture due date in statement.object.definition.extensions. (http://xapi.jisc.ac.uk/dueDate)

**Related Issues**
- LMS-2589 (action 1, action 3)

**PR Type**
Enhancement

### DIFF
--- a/src/transformer/utils/extensions/jisc.php
+++ b/src/transformer/utils/extensions/jisc.php
@@ -18,6 +18,7 @@ namespace src\transformer\utils\extensions;
 defined('MOODLE_INTERNAL') || die();
 
 use src\transformer\utils as utils;
+use src\transformer\utils\extensions as extensions;
 
 function jisc(array $config, \stdClass $event, $course) {
     if (utils\is_enabled_config($config, 'send_jisc_data')) {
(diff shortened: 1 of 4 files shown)


stephen-oh (approved):

src/transformer/utils/extensions/jisc.php
```diff
@@ -27,7 +28,7 @@ function jisc(array $config, \stdClass $event, $course) {
                 'http://id.tincanapi.com/extension/ip-address' => $event->ip,
                 'http://xapi.jisc.ac.uk/statementCat' => 'VLE',
             ],
-            course_area($course)
+            extensions\course_area($course)
         );
```
gordonmacqueen-lp: Can't we just do `utils\extensions\course_area` and drop the additional use statement on line 21?

lzabo: I added it to stay consistent with the style used in the rest of the plugin.

gordonmacqueen-lp: I've not found `use src\transformer\utils\extensions as extensions;` anywhere else in the codebase.

If we are going for consistency I think it should be `utils\extensions\course_area`, in a similar way it has been done in admin/tool/log/store/xapi/src/transformer/events/core/course_completed.php: 
`'object' => utils\get_activity\course($config, $course)`

Either that, or we drop the changes from this file (and base.php) altogether as they are not related to the ticket

gordonmacqueen-lp (approved):

PR #765, opened 2020-07-27
LMS-2940 logstore_xapi: add a day when datefrom and dateto is equal.

lzabo:
**Description**
Add an extra day to Dateto selector when Datefrom and Dateto is equal.

**Related Issues**
LMS-2940

**PR Type**
Fix

### DIFF
--- a/report.php
+++ b/report.php
@@ -171,16 +171,14 @@ if (isset($formelements)) {
     }
 
     if (!empty($formelements->datefrom)) {
-        if ($formelements->datefrom == $formelements->dateto) {
-            $formelements->dateto = $formelements->dateto + (DAYSECS - 1);
-        }
         $where[] = 'x.timecreated >= :datefrom';
         $params['datefrom'] = $formelements->datefrom;
     }
 
     if (!empty($formelements->dateto)) {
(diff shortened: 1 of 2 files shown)


report.php
```diff
@@ -171,6 +171,9 @@
     }
 
     if (!empty($formelements->datefrom)) {
+        if ($formelements->datefrom == $formelements->dateto) {
+            $formelements->dateto = $formelements->dateto + DAYSECS;
+        }
```
gordonmacqueen-lp: should this not be `$formelements->dateto + (DAYSECS - 1)`, so that you end up with 23:59:59 on the same date?

lzabo: I will change it.

gordonmacqueen-lp:
Perhaps we should change to a date time selector, to allow fine grained control
https://docs.moodle.org/dev/lib/formslib.php_Form_Definition#date_time_selector

From the spec (4.3):
> - A start date/time field, which will only return events that occurred on or after the specified date/time
> - An end date/time field, which will only return events that occurred on or before the specified date/time

gordonmacqueen-lp (approved):

PR #766, opened 2020-07-27
LMS-2881 - replace print_r() call with json_encode()

gordonmacqueen-lp:
**Description**
- replaces `print_r()` call with `json_encode()` to pass CI rules 

**PR Type**
- Fix

### DIFF
--- a/report.php
+++ b/report.php
@@ -268,7 +268,7 @@ if (!empty($results)) {
         if ($id == XAPI_REPORT_ID_ERROR) {
             $response = '';
             if (isset($result->response)) {
-                $response = '<pre>' . print_r(logstore_xapi_decode_response($result->response), true) . '</pre>';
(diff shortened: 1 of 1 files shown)


report.php
```diff
@@ -268,7 +268,7 @@
         if ($id == XAPI_REPORT_ID_ERROR) {
             $response = '';
             if (isset($result->response)) {
-                $response = '<pre>' . print_r(logstore_xapi_decode_response($result->response), true) . '</pre>';
+                $response = '<pre>' . json_encode(json_decode($result->response), JSON_PRETTY_PRINT)  . '</pre>';
```
lzabo: We should use logstore_xapi_decode_response instead of using json_decode or we can delete that function logstore_xapi_decode_response.

lzabo (changes requested):

report.php
gordonmacqueen-lp: I've updated that now @lzabo - didn't realise I had dropped it

lzabo (approved):

PR #799, opened 2021-05-27
Xapi store for moodle community

CinecaElearning:
**Description**
- added bestr data (firstname and lastname) to a new transformer extension named bestr

**PR Type**
- Enhancement


davidpesce:
This seems to be well outdated. Closing unless there's a desire to incorporate.

Issue #87, opened 2016-11-01
Adding custom user profile fields


ryasmi:
Question from [Dillon Esterhuizen](https://moodle.org/user/view.php?id=2041189&course=1) in [our Moodle comments](https://moodle.org/plugins/logstore_xapi). I've moved this here so that it's easier to manage and thread together the conversation.

> Hi there!
> 
> I have a similar question to Lee Kirkland who posted above. I'm trying to add custom user profile fields to the statements so that they show up in the LRS as well but I'm not sure how to add it in your code. I'm using the plugin version 1.0.0 and Learning Locker as the LRS.
> 
> I understand that there are 3 libraries I need to add it to, i.e. the expander, the translator and the emitter, but I can't figure out where inside of them to add it. I know that the standard log only records the user's id, which means that there must be some query somewhere that then gets all that user's information from the user table as well, but I can't seem to find it to add to that query.
> 
> Do I add it to the controller or the repo of each library, or both the controller and the repo (I would imagine that it shouldn't need to be added to any events as it is information required for every event)? And are there existing functions that I add it to or do I need to create a function specifically to query this data? Also, I would imagine I'd need to add it to the statements as well? I hope I'm asking the right questions.
> 
> Your help would be greatly appreciated!

ryasmi:
Thank you for the question Dillon. You're certainly asking the right questions and I hope this response helps.

### Changes to the Expander

If there are queries you need to make to the database, you will probably want to add a new method to the repository in the Expander. You will also need to change the [base Event class](https://github.com/LearningLocker/Moodle-Log-Expander/blob/master/src/Events/Event.php) so that it passes the additional data through to the Translator for all events.

### Changes to the Translator

You will also need to change the [base Event class](https://github.com/LearningLocker/Moodle-xAPI-Translator/blob/master/src/Events/Event.php) in the translator so that it passes the additional data through the Emitter.

### Changes to the Emitter

Again, you will need to change the [base Event class](https://github.com/LearningLocker/xAPI-Recipe-Emitter/blob/master/src/Events/Event.php) here so that the additional data is added to all statements and therefore sent to the LRS.

### That's a lot!

I know that's a lot of stuff to do and it's a pain to make the changes to three different repositories. Apologies for that, once I get more time on this project, I will hopefully be able to reduce the work required for changes of this nature.

dillonhubble:
Thanks @ryansmith94 !

[...]

PR #809, opened 2022-02-22
[Request] Added username and emailAddress to extensions

meetkelvincodes:
**Description**
- This PR is in reference to a few requests I noticed under issues that were closed, as well as a similar need for a project I am working on. The need is to be able to include an email address in addition to a username to the actor. Unfortunately since we cannot add to the actor schema, I noticed guidance in the comments that the requester should consider adding additional profile attributes to the extensions schema.  Reference: https://github.com/xAPI-vle/moodle-logstore_xapi/issues/87#issuecomment-257824618. 

- Please comment if there is an alternative option for including both username and email in the actor schema instead, without modifying the plugin to vastly.

**Considerations**
- Leverage Groups in the Actor Schema (Require larger modification of plugin to accomplish than enhancing extensions)
- Add to Extensions schema (Selected)
- Develop new schema (Large level of effort)
- Add to Actor schema (Can't due to restrictions)

**Related Issues**
- #87

**PR Type**
- Enhancement

**Tests**
- Not yet completed, waiting until confirmed solution is logical and sound by maintainers.
- Manually tested and confirmed working in the environment I'm actively supporting.

### DIFF
--- a/src/transformer/utils/extensions/info.php
+++ b/src/transformer/utils/extensions/info.php
@@ -18,12 +18,22 @@ namespace src\transformer\utils\extensions;
 defined('MOODLE_INTERNAL') || die();
 
 function info(array $config, \stdClass $event) {
+    $repo = $config['repo'];
+    $actorId = property_exists($event, 'relateduserid') &&
+                isset($event->relateduserid) ? $event->relateduserid:$event->userid;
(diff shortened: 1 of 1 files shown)


davidpesce:
This feature was implemented in the prior work done by YET

PR #820, opened 2022-10-07
Master jisc

ryasmi:
@gordonmacqueen-lp just looking at the diff here.

CC: @davidpesce


davidpesce:
@ryansmith94 @gordonmacqueen-lp I'm OK with this. I've got a bunch of work completed already on the master-jisc to bring this closer to automated release (and all checks green). The jisc-data checkbox is already in the master branch and isn't impacting users right now. There are also quite a few forked repos that have continued on without us. I've reached out to those individuals and they're eager to get this updated.

davidpesce:
Oh, and the concerns on the performance issues with a large Moodle logstore... this is a common problem with Moodle and unrelated to this plugin. There are ways to mitigate this by offloading the logstore to things like s3, etc.

davidpesce (approved):

PR #835, opened 2023-02-16
File for pseudonymisation

AD1529:
Hi,

These are the files that we updated to allow the user to pseudonymise any data that could be used to identify them.

Best,
Daniela


AD1529:
I updated some files

PR #836, opened 2023-02-18
Pseudonimisation and role assigned 

AD1529:
Hi,

I updated some files.


davidpesce:
Thanks for contributing this; it seems like quite a bit of work!

It's missing the unit tests for each of the new events being monitored. Can you add those? If you are looking for an example, check out the tests folder in this repository. 

In an ideal world, each new feature would be it's own PR. Is it possible to send them as separate PRs?

AD1529:
I'll open a new one

PR #840, opened 2023-04-20
All the work done.

AD1529:
**Description**
Hi David, here you can find all the statements and the improvement to the plugin for pseudonymisation and the handle of statements in Moodle interface.

**Related Issues**
Other work should be done in the future to further improve statements. But this can be done by analysing the results of the extraction of data from the LRS.

**PR Type**
- Enhancement


AD1529:
I'm uploading a new version

PR #855, opened 2024-05-06
timezone patch

CinecaElearning:
**Description**
- sets back the default_timezone value for other moodle plugins

**PR Type**
- Fix


davidpesce:
Thank you for contributing this! We've switched to using UTC exclusively, so this is no longer needed.

PR #856, opened 2024-05-06
bestr data added to the statement

CinecaElearning:
**Description**
- bestr data extension. adds birthdate and codice fiscale (CF) (an italian unique personal identifier similar to the well known social security number).

**PR Type**
- Enhancement


davidpesce:
This very specific to the Bestr platform and I don't think it should be included in a plugin like this.

PR #861, opened 2024-12-12
Yet Logstore Work 2024

milt:
**Description**
This PR:
* adds ~100 new events
* makes numerous improvements to existing events
* makes numerous test improvements to reduce boilerplate

**PR Type**
- Fix, Enhancement


davidpesce:
Hi Milt - first off, thank you for this enormous effort. It looks awesome! It's going to take me a bit of time to review, but I'm curious why the CI was removed from pull requests. Could you re-enable that in this PR?

milt:
> Hi Milt - first off, thank you for this enormous effort. It looks awesome! It's going to take me a bit of time to review, but I'm curious why the CI was removed from pull requests. Could you re-enable that in this PR?

Hi David! As far as I could tell running CI on every push covers pull requests as well, but I'm happy to turn it back on.

milt:
We added PHP 8.3 compatibility and some other minor fixes. The support range is now PHP 8.1 to 8.3 and moodle 4.1 to 4.5.

src/loader/moodle_curl_lrs.php
```diff
@@ -53,19 +53,39 @@ function load(array $config, array $events) {
...
         $postdata = json_encode($statements);
 
+
+
         if ($postdata === false) {
             throw new \Exception('JSON encode error: '.json_last_error_msg());
         }
 
-        $request = new \curl();
+        //Old Client Code
```
davidpesce: Can these comments and debug be removed?

src/transformer/events/core/badge_awarded.php
```diff
@@ -0,0 +1,104 @@
...
+        'context' => [
+            ...utils\get_context_base($config, $event, $lang, $course),
+            'instructor' => $awarder,
+            'contextActivities' =>  [
+                'category' => [[
+                    'id' => $config['app_url'],
+                    'objectType' => 'Activity',
+                    'definition' => [
+                        'name' => [
+                            'en' => 'EDLM Moodle LMS'
```
davidpesce: Can this be made agnostic? ie "Moodle LMS"

davidpesce: It's all over, so maybe a global search and replace is better.

src/transformer/utils/get_activity/certificate.php
davidpesce, line 1: I'm curious why this was renamed rather than replaced?

davidpesce (changes requested):
This is overall awesome! Thank you so much for contributing! 

Please take a look at the comments and let me know your thoughts.

src/transformer/events/core/badge_awarded.php
milt: We actually refactored out the LMS category activity but apparently this file didn't get the memo ;). I'll make sure they all use the same fn. As for the name value I can make it generic, but perhaps we should add config for it?

src/transformer/utils/get_activity/certificate.php
milt: This _appears_ to be git trying to be cleverer than it should. We removed the original file after centralizing all course module object generation to a single function. Separately we added this util file for generating objects representing tool_certificate certificates. Since there is so much boiler in between the two files it looks like git sees that as a rename?

src/transformer/events/core/badge_awarded.php
davidpesce: Hmmm, yeah, an admin setting would be awesome!

milt: So on further examination it appears to use the full name from the ur-course if available.  https://github.com/yetanalytics/moodle-logstore_xapi/blob/master/src/transformer/utils/get_activity/site.php#L37-L49
I'm gonna see if the user experience for setting that makes sense, but also make the default "Moodle LMS" and fix those badge handlers.

milt: Yep, the UI exposes it in site home settings, so I think we're good:
image

milt: OK, this work surfaced an issue with our testing, since Moodle stores information on the hidden course ID 1, we should really be using an ID of 2 for all the actual courses in our test data rather than 1. I'm gonna fix that too.

milt: OK! This is under internal review but should be updated on this PR soon https://github.com/yetanalytics/moodle-logstore_xapi/pull/77

milt:
@cliffcaseyyet noticed that there doesn't seem to be a way to set the statement `$.context.platform` from the UI. It currently derives platform (which we added) from the `source_name` config but I don't think that has UI and seems to be hardcoded in the config. I'm gonna make another PR on our side to add a new config for platform so you can, for instance, disambiguate between multiple peer instances of Moodle.

src/loader/moodle_curl_lrs.php
cliffcaseyyet: this should be resolved.

milt:
OK @davidpesce it should be all caught up with our changes now. Let us know if you find anything else.

davidpesce (approved):

Issue #872, opened 2025-09-22
PHP 8.3 compatibility


sharpchi:
Running PHPStan has revealed a few "property.notFound" errors which can be fixed by defining them in the class.

```
------ ------------------------------------------------------------------------------------------------------ 
  Line   web/html/admin/tool/log/store/xapi/classes/log/moveback.php                                           
 ------ ------------------------------------------------------------------------------------------------------ 
  :86    Access to an undefined property logstore_xapi\log\moveback::$type.                                    
         🪪  property.notFound                                                                                  
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property                   
  :124   Access to an undefined property logstore_xapi\log\moveback::$type.                                    
         🪪  property.notFound                                                                                  
         💡 Learn more: https://phpstan.org/blog/solving-phpstan-access-to-undefined-property                   
 ------ ------------------------------------------------------------------------------------------------------
````

davidpesce:
Thanks for the PR. This has been merged.

PR #875, opened 2025-09-22
Import type not defined

sharpchi:
**Description**
- PHP 8.3 requires class vars to be defined

**Related Issues**
- #872 

**PR Type**
- Fix


davidpesce:
Thanks for this, Mark! Looks good.

PR #890, opened 2026-03-04
Support events for mod_game

davidherney:
**Description**
- mod_game event handling. The Game plugin is widely used on Moodle platforms.

**Related Issues**
- NA

**PR Type**
- Enhancement


davidpesce:
Hi @davidherney, thanks for this PR! I haven't had a chance to dive into it too deeply, but could you remove your changes to version.php from the PR. It's better to handle version bumps from our end.

davidherney:
Hola @davidpesce 

Done, I reverted the change to version.php.

Saludos

src/transformer/get_event_function_map.php
```diff
@@ -126,6 +126,9 @@ function get_event_function_map() {
         '\mod_forum\event\discussion_subscription_created' => 'mod_forum\discussion_subscription_created',
         '\mod_forum\event\discussion_subscription_deleted' => 'mod_forum\discussion_subscription_deleted',
         '\mod_forum\event\user_report_viewed' => 'mod_forum\user_report_viewed',
+        '\mod_game\event\course_module_instance_list_viewed' => 'mod_game\course_module_instance_list_viewed',
+        '\mod_game\event\course_module_viewed' => 'mod_game\course_module_viewed',
```
davidpesce: **course_module_viewed should use the shared transformer **

  Every other module maps course_module_viewed to all\course_module_viewed:
  '\mod_assign\event\course_module_viewed' => 'all\course_module_viewed',
  '\mod_quiz\event\course_module_viewed'   => 'all\course_module_viewed',
  '\mod_forum\event\course_module_viewed'  => 'all\course_module_viewed',
  ... (22 modules total)

  This PR creates a redundant custom mod_game\course_module_viewed that does the same thing. The map entry should be:
  '\mod_game\event\course_module_viewed' => 'all\course_module_viewed',

  The custom transformer file (src/transformer/events/mod_game/course_module_viewed.php) and its test directory can be removed entirely — the existing
  all\course_module_viewed test coverage already validates this pattern.

src/transformer/events/mod_game/course_module_instance_list_viewed.php
```diff
@@ -0,0 +1,72 @@
...
+                'en' => 'Viewed',
+            ],
+        ],
+        'object' => [
+            'id' => $config['app_url'] . '/mod/game/index.php?id=' . $event->courseid,
+            'objectType' => 'Activity',
+            'definition' => [
+                'type' => 'http://id.tincanapi.com/activitytype/collection-simple',
+                'name' => [
+                    $lang => 'Game instances list',
```
davidpesce, line 56: Should this be pulled from the lang file of mod_game?

davidpesce (changes requested):
I think we're good after these two fixes.

src/transformer/get_event_function_map.php
davidherney: You're right. I already removed the duplicate and handled it with all\course_module_viewed

`'\mod_game\event\course_module_viewed' => 'all\course_module_viewed',`

src/transformer/events/mod_game/course_module_instance_list_viewed.php
davidherney: Hi @davidpesce 

I don't manage the **mod_game** component, and it seems unnecessary to send a pull request to the owner for an informational string, especially since many Transformers use it the same way. For example:
```
                'name' => [
                    $lang => 'Global Search',
                ],
```
```
            'definition' => [
                'type' => 'http://adlnet.gov/expapi/activities/file',
                'name' => [
                    $lang => 'Questions in ' . $info['format'] . ' format',
                ],
            ],
```

I could also change it to a generic **mod_game** string, but the closest one used as a title in the Games listing is:
`$string['modulenameplural'] = 'Games';`

How would you prefer I proceed?

Saludos

